PR #24, opened 2021-08-30
remove intermediate static lib

mattnite:

Hey! I ran into linking issues when incorporating this into a project because the engine lib was picking up my own `src/main.zig`, I don't think this intermediate static lib is actually needed (at least on linux) so adding source directly to the step should do. (fingers crossed on that CI)

### DIFF
--- a/glfw/build.zig
+++ b/glfw/build.zig
@@ -40,10 +40,6 @@ pub fn link(b: *Builder, step: *std.build.LibExeObjStep, options: Options) void 
     var arena = std.heap.ArenaAllocator.init(b.allocator);
     defer arena.deinit();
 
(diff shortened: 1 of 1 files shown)


slimsag:
You rock @mattnite :) This looks great and makes sense.

It seems `addStaticLibrary` is not relative to build.zig, same issue we've run into with adding C sources (the reason we make those paths absolute / relative to build.zig first.) I guess we could keep the static lib, which might help with compile times in the future, if we did the same thing with making the `src/main.zig` include relative to our build.zig.

For now, though, this seems good enough! @Avokadoen I suspect this is the same issue you were experiencing when you sent #23 so maybe you can switch back to upstream soon :)

Avokadoen:
Good job @mattnite and @slimsag! 😄

PR #62, opened 2021-11-05
gpu: beginnings of Dawn WebGPU backend

slimsag:
What works:

* https://github.com/hexops/dawn is a fork of Google's Dawn which will track upstream closely, only difference is it has generated code committed so we do not require installation of `depot_tools`, `ninja`, `make`, `gn`, etc. tools that Google requires.
* `build.zig` builds all of Dawn using only the Zig compiler, currently only working on MacOS but should be easy to get building on other targets soon. Depends on PR I sent ziglang/zig#10096 to enable Zig to compile Obj-C++ code.
* `webgpu/src/dawn` contains C shims for all the dawn_native functionality which is required to get Dawn actually attached to a window / set up.
* `cd webgpu && zig build run-dawn-example` shows it all working:

This PR does not enable cross-compilation of Dawn. It does not add Zig WebGPU wrappers. These will come later.

- [x] By selecting this checkbox, I agree to license my contributions to this project under the license(s) described in the LICENSE file, and I have the right to do so or have received permission to do so by an employer or client I am producing work for whom has this right.

PR #63, opened 2021-11-05
glfw: fix linking difference between sysroot and non-sysroot

slimsag:
This effectively gives us the dependencies we need in any case, and works around ziglang/zig#10103

Importantly, this removes a blocker for landing WebGPU support in hexops/mach#62


slimsag:
Strange this works on both x86_64-macos and aarch64-macos, but not windows/linux cross compilation.

Linux complains of missing: CoreGraphics, Foundation

This may be due to our MacOS system SDK being older than what is used on the Mac CI runners. For now will add a check for non-Mac hosts to link these.

PR #67, opened 2021-11-10
Friendlier GLFW API initialization

### LINKED ISSUE #66: Friendlier GLFW API initialization
> We should also do what this issue proposed for `glfw.Init`, as there are init hints for GLFW (and potentially one other type of hint? I forget off the top of my head)

_Originally posted by @slimsag in https://github.com/hexops/mach/issues/65#issuecomment-963626820_

InKryption:
Closes #66

What should be done with the original `InitHint` enum and `initHint` function could maybe be discussed. I didn't remove them, as they were useful for implementing the new interface, and make it so that new code behaves about the same way as old code which explicitly set all hints. They are also documentative, seeing as they more closely reflect the actual GLFW API than these changes. 

### DIFF
--- a/glfw/README.md
+++ b/glfw/README.md
@@ -67,7 +67,7 @@ Now in your code you may import and use GLFW:
 const glfw = @import("glfw");
 
 pub fn main() !void {
-    try glfw.init();
+    try glfw.init(.{});
     defer glfw.terminate();
 
     // Create our window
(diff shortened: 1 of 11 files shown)


slimsag (approved):
This looks really great, thank you for doing this!

I like how you kept the current initHint function and enum, that will also make it easier to update GLFW when new changes are made upstream.

InKryption:
> I like how you kept the current initHint function and enum, that will also make it easier to update GLFW when new changes are made upstream.

Good, I'm glad we're on the same page.

PR #68, opened 2021-11-11
glfw: remove [*c] pointers from api and return lengthed strings when possible

silversquirl:
Just makes the API a bit nicer to use from Zig. According to the [Zig docs](https://ziglang.org/documentation/master/#toc-C-Pointers) `[*c]` pointers should be avoided, and should be replaced with more precise pointer types wherever possible

### DIFF
--- a/glfw/src/Joystick.zig
+++ b/glfw/src/Joystick.zig
@@ -217,10 +217,13 @@ pub inline fn getHats(self: Joystick) Error!?[]const Hat {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: joystick_name
-pub inline fn getName(self: Joystick) Error![*c]const u8 {
-    const name = c.glfwGetJoystickName(self.jid);
+pub inline fn getName(self: Joystick) Error!?[:0]const u8 {
+    const name_opt = c.glfwGetJoystickName(self.jid);
     try getError();
-    return name;
+    return if (name_opt) |name|
+        std.mem.span(name)
+    else
+        null;
 }
 
 /// Returns the SDL compatible GUID of the specified joystick.
@@ -250,10 +253,13 @@ pub inline fn getName(self: Joystick) Error![*c]const u8 {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: gamepad
-pub inline fn getGUID(self: Joystick) Error![*c]const u8 {
-    const guid = c.glfwGetJoystickGUID(self.jid);
+pub inline fn getGUID(self: Joystick) Error!?[:0]const u8 {
+    const guid_opt = c.glfwGetJoystickGUID(self.jid);
     try getError();
-    return guid;
+    return if (guid_opt) |guid|
+        std.mem.span(guid)
+    else
+        null;
 }
 
 /// Sets the user pointer of the specified joystick.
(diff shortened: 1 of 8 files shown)


InKryption:
Would you consider omitting the usage of `std.mem.span`, and just return a sentinel-terminated pointer? It's probably not significant, but it is extra overhead that can be avoided by leaving the responsibility of using `std.mem.span` on the pointer to the user. If one of these slices are ever forwarded to any other C api for any reason, then the length has to be omitted, and the C api will calculate the length again.

silversquirl:
imo a couple extra nanoseconds per call is worth it for the nicer API. I doubt you'd even notice the difference if you benchmarked it. Maybe I'm missing something though

glfw/src/Joystick.zig
```diff
@@ -407,10 +413,13 @@ pub inline fn isGamepad(self: Joystick) bool {
...
-pub inline fn getGamepadName(self: Joystick) Error!?[*c]const u8 {
-    const name = c.glfwGetGamepadName(self.jid);
+pub inline fn getGamepadName(self: Joystick) Error!?[:0]const u8 {
+    const name_opt = c.glfwGetGamepadName(self.jid);
     try getError();
-    return name;
+    return if (name_opt) |name|
+        std.mem.span(name)
+    else
+        null;
```
slimsag, line 422: If I understand correctly, `std.mem.span` handles optional pointer types already so we may not need these null checks? https://github.com/ziglang/zig/blob/33889e1974137a54d1a88b87f57462de2a7cc145/lib/std/mem.zig#L623-L629

glfw/src/clipboard.zig
```diff
@@ -39,10 +39,10 @@ pub inline fn setClipboardString(value: [*c]const u8) Error!void {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: clipboard, glfwSetClipboardString
-pub inline fn getClipboardString() Error![*c]const u8 {
+pub inline fn getClipboardString() Error![:0]const u8 {
     const value = c.glfwGetClipboardString(null);
     try getError();
-    return value;
+    return std.mem.span(value);
```
slimsag, line 45: If we do need the null checks, do we not also need one here?

slimsag:
This looks great! I left a question about the null checks however. Once that is settled I'll merge this.

glfw/src/Joystick.zig
silversquirl: Good point, I should've checked that :)

silversquirl: Ah, I remember why this was. `*c` pointers are not optional types, so span won't null-check them

glfw/src/clipboard.zig
silversquirl: We don't, because that function will never return null according to GLFW docs

slimsag: Ah, okay, I see now it only returns NULL if an error occurred. Makes sense!

glfw/src/Joystick.zig
slimsag: Ahh got it!

slimsag (approved):
Thanks for this!

silversquirl:
Replaced your merge with a rebase for slightly cleaner history :)
Should be good to merge into upstream now

slimsag:
Awesome, thank you!

PR #70, opened 2021-11-11
glfw: Small Window fixes

### LINKED ISSUE #69: glfw.Window.Size should be public
It's returned by public APIs (`glfw.Window.getSize` etc), so it's inconvenient to have the type be private.

InKryption:
Closes #69, and applies the same thought to other parts of the API. Also fixes some stray comments which should be doc comments. 


slimsag (approved):
Nice job catching these, I appreciate it!

Issue #65, opened 2021-11-08
glfw: Friendlier window creation


silversquirl:
GLFW's C API uses "window hints", which aren't hugely user friendly. In Zig, we can do much better by using an "options" struct with default values. My own GLFW wrapper [does this](https://github.com/silversquirl/glfz/blob/main/types.zig#L16) for GLFW's window creation, which makes the API much nicer in my opinion.

It could be argued this deviates too far from the C API, but at the same time it doesn't really change much - just combines a few function calls into one, making the API use Zig idioms. I think of it similarly to error handling.

Here's an example of how window creation could look using an options struct:
```zig
const window = try glfw.Window.create(640, 480, "Hello, mach-glfw!", .{
    .monitor = monitor,
    .context_version_major = 4,
    .context_version_minor = 5,
    .opengl_profile = .core,
});
```

InKryption:
I think this ties in nicely to what #61 is trying to accomplish, by removing the redundancy of access to `glfw.Window.hint` after window creation.

Avokadoen:
I think if we get a couple more of these types of ideas, then it would be worth creating a glfw repo with a more zig friendly API and keep the current repo as is. There are of course some "zigified" concepts already in the project i.e changes related to issue #37 so the trade-off between QoL and too much of a deviation is hard to define currently. Issue #61 and this issue does change the overall "look" of user code so I guess that kinda indicate a stronger deviation.

Maybe @slimsag could simply create a tag for "ziggify glfw" or whatever that indicates a issue that can be closed. Then we can at a later stage (or whenever someone feels like taking the leap) easily find all these issues to implement them

slimsag:
Honestly, it seems like there is a desire for this from many people. And if there's enough desire that people would want to create another library on top, it seems reasonable to me to just do it here. If people do want the true C GLFW API, they can always use both together.

Maybe we just dive all-in and accept this proposal as well as #61 ?

slimsag:
I should also maybe note, there will later be an application library exposed by Mach specifically for setting up WebGPU applications which will be a wrapper over GLFW, HTML5/WebAssembly, Android and iOS APIs. Not sure how much it would help people going for a Vulkan or OpenGL-only approach, though.

Avokadoen:
The one caveat I think is truly important if such changes become accepted is that they need to be _clearly_ documented. And the readme should warn about there being deviations in the API

slimsag:
Yes, I agree with that. Seems easy enough to do though :)

[...]

PR #71, opened 2021-11-11
glfw: Window hints rework

InKryption:
~~I took a different approach to the one used for initialization hints, because I realized that it would probably a lot of overhead to call `Window.hint` for every possible hint, compared to how status quo allows one to avoid that overhead; as well, it would probably be a lot harder to maintain, given that, unlike `InitHint`, window initialization hints are of more types than just booleans.~~
~~This approach is (quite a bit) more verbose, but avoids the overhead, and probably makes it easier to maintain.~~

Also, I found that `context_revision` and `context_no_error` are listed in `Window.Hint`, but are not listed as hints in the docs, only as attributes; I commented them out to make what I have pass tests - is this intentional and I'm missing something, or is this the result of some mishap? 

Edit: ~~it should also be noted that unlike the change made to initialization hints, I didn't completely re-use the in-place API, because I ran into some issues regarding the type-inference in the `Window.hint` function, so you'll see I ended up re-implementing the code for `Window.hint` in `HintValue.set` (where `HintValue` is a tagged union, with a tag type of `Hint`).~~

Edit: as pointed out by @silversquirl, there isn't actually a lot of overhead involved in the struct approach, so now the union slice is no more. ~~Now just to address the other components of this problem, such as `Window.hint`, which seems to have some some type inference issues which caused compilation to fail under this new model. Should we try and rework it, or discard it in favor of `Hints.set` (private function)?~~

### DIFF
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -97,7 +97,7 @@ pub const Hint = enum(c_int) {
     maximized = c.GLFW_MAXIMIZED,
 
     /// Cursor centering window hint
-    center = c.GLFW_CENTER_CURSOR,
+    center_cursor = c.GLFW_CENTER_CURSOR,
 
     /// Window framebuffer transparency hint
     transparent_framebuffer = c.GLFW_TRANSPARENT_FRAMEBUFFER,
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -78,7 +78,7 @@ pub const Hint = enum(c_int) {
     /// Input focus window hint.
     focused = c.GLFW_FOCUSED,
 
-    // Window resize-ability window hint
+    /// Window resize-ability window hint
     resizable = c.GLFW_RESIZABLE,
 
     /// Window visibility window hint
@@ -1048,7 +1048,7 @@ pub const Attrib = enum(c_int) {
     /// Window iconification window attribute.
     iconified = c.GLFW_ICONIFIED,
 
-    // Window resize-ability window attribute
+    /// Window resize-ability window attribute
     resizable = c.GLFW_RESIZABLE,
 
     /// Window visibility window attribute
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -510,7 +510,7 @@ pub inline fn setIcon(self: Window, allocator: *mem.Allocator, images: ?[]Image)
     try getError();
 }
 
-const Pos = struct {
+pub const Pos = struct {
     x: usize,
     y: usize,
 };
@@ -561,7 +561,7 @@ pub inline fn setPos(self: Window, pos: Pos) Error!void {
     try getError();
 }
 
-const Size = struct {
+pub const Size = struct {
     width: usize,
     height: usize,
 };
@@ -693,7 +693,7 @@ pub inline fn getFramebufferSize(self: Window) Error!Size {
     return Size{ .width = @intCast(usize, width), .height = @intCast(usize, height) };
 }
 
-const FrameSize = struct {
+pub const FrameSize = struct {
     left: usize,
     top: usize,
     right: usize,
@@ -1709,7 +1709,7 @@ pub inline fn getMouseButton(self: Window, button: MouseButton) Error!Action {
     return @intToEnum(Action, state);
 }
 
-const CursorPos = struct {
+pub const CursorPos = struct {
     xpos: f64,
     ypos: f64,
 };
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -75,117 +75,61 @@ inline fn defaultHints() Error!void {
 }
 
 /// Window hints
-pub const Hint = enum(c_int) {
-    /// Window resize-ability window hint
+const Hint = enum(c_int) {
     resizable = c.GLFW_RESIZABLE,
-
-    /// Window visibility window hint
     visible = c.GLFW_VISIBLE,
-
-    /// Window decoration window hint
     decorated = c.GLFW_DECORATED,
-
-    /// Input focus window hint.
     focused = c.GLFW_FOCUSED,
-
-    /// Window auto-iconification window hint
     auto_iconify = c.GLFW_AUTO_ICONIFY,
-
-    /// Window decoration window hint
     floating = c.GLFW_FLOATING,
-
-    /// Window maximization window hint
     maximized = c.GLFW_MAXIMIZED,
-
-    /// Cursor centering window hint
     center_cursor = c.GLFW_CENTER_CURSOR,
-
-    /// Window framebuffer transparency hint
     transparent_framebuffer = c.GLFW_TRANSPARENT_FRAMEBUFFER,
-
-    /// Input focus on calling show window hint
     focus_on_show = c.GLFW_FOCUS_ON_SHOW,
+    scale_to_monitor = c.GLFW_SCALE_TO_MONITOR,
 
-    /// Framebuffer bit depth hint.
+    /// Framebuffer hints
     red_bits = c.GLFW_RED_BITS,
-
-    /// Framebuffer bit depth hint.
     green_bits = c.GLFW_GREEN_BITS,
-
-    /// Framebuffer bit depth hint.
     blue_bits = c.GLFW_BLUE_BITS,
-
-    /// Framebuffer bit depth hint.
     alpha_bits = c.GLFW_ALPHA_BITS,
-
-    /// Framebuffer bit depth hint.
     depth_bits = c.GLFW_DEPTH_BITS,
-
-    /// Framebuffer bit depth hint.
     stencil_bits = c.GLFW_STENCIL_BITS,
-
-    /// Framebuffer bit depth hint.
     accum_red_bits = c.GLFW_ACCUM_RED_BITS,
-
-    /// Framebuffer bit depth hint.
     accum_green_bits = c.GLFW_ACCUM_GREEN_BITS,
-
-    /// Framebuffer bit depth hint.
     accum_blue_bits = c.GLFW_ACCUM_BLUE_BITS,
-
-    /// Framebuffer bit depth hint.
     accum_alpha_bits = c.GLFW_ACCUM_ALPHA_BITS,
-
-    /// Framebuffer auxiliary buffer hint.
     aux_buffers = c.GLFW_AUX_BUFFERS,
-
-    /// OpenGL stereoscopic rendering hint.
-    stereo = c.GLFW_STEREO,
-
-    /// Framebuffer MSAA samples hint.
+    
+    /// Framebuffer MSAA samples
     samples = c.GLFW_SAMPLES,
-
-    /// Framebuffer sRGB hint.
-    srgb_capable = c.GLFW_SRGB_CAPABLE,
-
-    /// Monitor refresh rate hint.
+    
+    /// Monitor refresh rate
     refresh_rate = c.GLFW_REFRESH_RATE,
 
-    /// Framebuffer double buffering hint.
+    /// OpenGL stereoscopic rendering
+    stereo = c.GLFW_STEREO,
+
+    /// Framebuffer sRGB
+    srgb_capable = c.GLFW_SRGB_CAPABLE,
+
+    /// Framebuffer double buffering
     doublebuffer = c.GLFW_DOUBLEBUFFER,
 
-    /// Context client API hint.
     client_api = c.GLFW_CLIENT_API,
-
-    /// Context client API major version hint.
-    context_version_major = c.GLFW_CONTEXT_VERSION_MAJOR,
-
-    /// Context client API minor version hint.
-    context_version_minor = c.GLFW_CONTEXT_VERSION_MINOR,
-
-    /// Context robustness hint.
-    context_robustness = c.GLFW_CONTEXT_ROBUSTNESS,
-
-    /// OpenGL forward-compatibility hint.
-    opengl_forward_compat = c.GLFW_OPENGL_FORWARD_COMPAT,
-
-    /// Debug mode context hint.
-    opengl_debug_context = c.GLFW_OPENGL_DEBUG_CONTEXT,
-
-    /// OpenGL profile hint.
-    opengl_profile = c.GLFW_OPENGL_PROFILE,
-
-    /// Context flush-on-release hint.
-    context_release_behavior = c.GLFW_CONTEXT_RELEASE_BEHAVIOR,
-
-    /// Context error suppression hint.
-    context_no_error = c.GLFW_CONTEXT_NO_ERROR,
-
-    /// Context creation API hint.
     context_creation_api = c.GLFW_CONTEXT_CREATION_API,
 
-    /// Window content area scaling window
-    scale_to_monitor = c.GLFW_SCALE_TO_MONITOR,
+    context_version_major = c.GLFW_CONTEXT_VERSION_MAJOR,
+    context_version_minor = c.GLFW_CONTEXT_VERSION_MINOR,
+
+    context_robustness = c.GLFW_CONTEXT_ROBUSTNESS,
+    context_release_behavior = c.GLFW_CONTEXT_RELEASE_BEHAVIOR,
+    context_no_error = c.GLFW_CONTEXT_NO_ERROR,
+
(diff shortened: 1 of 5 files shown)


silversquirl:
> it would probably a lot of overhead to call Window.hint for every possible hint

I had this thought initially too, but if you look at the [GLFW source](https://github.com/glfw/glfw/blob/master/src/window.c#L255-L426), it's probably about the same as calling `glfwDefaultWindowHints`. I'd say have a hints struct with default values, and just set all of them unconditionally.

If you do want to avoid setting hints that don't need changed though, I'd still recommend going the struct route, and just check if the value is different than the default. This is easy to do by just looping over `std.meta.fields`, since the typeinfo gives you default values.

A struct results in a nicer API than an array of hints - the latter still feels very much like a C API, not a Zig one.

InKryption:
@silversquirl Good point. I don't know why I didn't actually check the source; guess that's what I get for starting this on my laptop at a café. And I guess that's also why I made this a draft.

InKryption:
@silversquirl Do you have any thoughts on what to do with the currently discarded `Window.hint` function?

silversquirl:
Either move the type conversion logic in `set` into it, or delete it. Not much point in keeping it if it's not used

glfw/src/Window.zig
```diff
@@ -266,6 +271,120 @@ pub inline fn hint(h: Hint, value: anytype) Error!void {
...
+                c_int => c.glfwWindowHint(hint_tag, hint_value),
+                
+                ClientApi,
+                ContextCreationApi,
+                ContextRobustness,
+                ContextReleaseBehavior,
+                OpenGlProfile,
+                => c.glfwWindowHint(hint_tag, @enumToInt(hint_value)),
+                
+                ([:0]const u8) => c.glfwWindowHintString(hint_tag, hint_value.ptr),
```
silversquirl, line 374: Any particular reason for these parens? Personally I find it more readable without them

InKryption: My personal reason for it is that I don't like "mixing" whitespace "scopes" (the obligatory space between 'const' and 'u8', and the optional whitespace between 'u8' and '=>'), if that makes any sense. Though, I have no particular attachment to it either, just a sort of typing habit. I can go ahead and change it whenever I get on my laptop.

InKryption:
Excuse my fumbling with git, trying to make the history not look messy.

glfw/src/Window.zig
```diff
@@ -162,8 +163,10 @@ pub const Hint = enum(c_int) {
     /// Context client API minor version hint.
     context_version_minor = c.GLFW_CONTEXT_VERSION_MINOR,
 
-    /// Context client API revision number hint.
-    context_revision = c.GLFW_CONTEXT_REVISION,
+    // TODO: The docs don't specify GLFW_CONTEXT_REVISION as a hint, only as an attribute; is there a reason this is here? Commented to make current
+    // draft work, which is mostly copying from the docs
```
silversquirl, line 167: The source backs you up here. GLFW_CONTEXT_REVISION is not a hint

glfw/src/Window.zig
```diff
@@ -180,8 +183,10 @@ pub const Hint = enum(c_int) {
     /// Context flush-on-release hint.
     context_release_behavior = c.GLFW_CONTEXT_RELEASE_BEHAVIOR,
 
-    /// Context error suppression hint.
-    context_no_error = c.GLFW_CONTEXT_NO_ERROR,
+    // TODO: The docs don't specify GLFW_CONTEXT_REVISION as a hint, only as an attribute; is there a reason this is here? Commented to make current
+    // draft work, which is mostly copying from the docs
```
silversquirl, line 187: This one is a hint, and is actually [listed in the docs](https://www.glfw.org/docs/latest/window_guide.html#GLFW_CONTEXT_NO_ERROR_hint)

glfw/src/Window.zig
```diff
@@ -205,66 +210,119 @@ pub const Hint = enum(c_int) {
...
+        any_release_behavior = consts.any_release_behavior,
+        release_behavior_flush = consts.release_behavior_flush,
+        release_behavior_none = consts.release_behavior_none,
+    };
+    
+    pub const OpenGlProfile = enum(c_int) {
+        opengl_any_profile = consts.opengl_any_profile,
+        opengl_compat_profile = consts.opengl_compat_profile,
+        opengl_core_profile = consts.opengl_core_profile,
+    };
```
silversquirl, line 296: Personally I'd remove these constants from consts and inline them here. They don't need to be publicly accessible any more, so there's probably no reason for them to be in consts.

glfw/src/Window.zig
InKryption: @silversquirl Is that the same case for `context_no_error = c.GLFW_CONTEXT_NO_ERROR`?

glfw/src/Window.zig
InKryption: @silversquirl I was thinking the same thing, but was unsure whether to do so being that something like this wasn't done in #56 already.

glfw/src/Window.zig
InKryption: @silversquirl Right, I didn't see that. I was looking [at this list](https://www.glfw.org/docs/latest/window_guide.html#window_hints_values), which lists hints, their default values and possible values. It's not in that list, and I can't see anywhere else that would specify a default value for it; maybe it could default to `!std.debug.runtime_safety`? (set to true in release modes).

silversquirl: Since we handle errors through Zig error unions, I'd say definitely default it to false. Possibly even remove it from the struct altogether, since it's quite an unsafe option to use - at the very least put a doc comment informing about the dangers.

InKryption: @silversquirl Alright, I suppose that makes sense. I'll amend it now and leave this as resolved.

glfw/src/Window.zig
InKryption: @silversquirl I went ahead with the change; the only constant in consts.zig now is `dont_care`.

slimsag:
Thanks for all your work on this! It'll probably be a few days before I have a chance to review (at handmade Seattle and then to a wedding) so apologies for the delayed review!

InKryption:
I had an idea to maybe add a test that compares the attribute values of a window created with `Window.from(c.glfwCreateWindow(...))`, to the attribute values of a window created with `Window.create(...)` (in that order), to check that, for any attribute whose value is initialized via `Window.Hints.set`, its value is the same as what would be set by GLFW itself (maybe excluding context_no_error). This wouldn't be full-coverage, but it would be something. Thoughts?

glfw/src/Window.zig
slimsag: Happy to have this as-is and let `zig fmt` formally sort it out later.

slimsag:
@InKryption regarding:

> I had an idea to maybe add a test that compares the attribute values of a window

This sounds like an obvious win to me, so yes, please do! I won't block this PR on this, please send another PR for it if I merge before you get to it.

InKryption:
@slimsag Roger roger, I'll set this as ready for merging then, in the absence of any notes you have.

glfw/src/Window.zig
```diff
@@ -162,9 +163,6 @@ pub const Hint = enum(c_int) {
     /// Context client API minor version hint.
     context_version_minor = c.GLFW_CONTEXT_VERSION_MINOR,
 
-    /// Context client API revision number hint.
-    context_revision = c.GLFW_CONTEXT_REVISION,
```
slimsag, line 166: The reason I had this here was that it is listed in the source as being both a hint and attribute:

https://github.com/glfw/glfw/blob/fb0f2f92a38c1d6a776ffeb253329f8d1c65694c/include/GLFW/glfw3.h#L1031-L1036

However, indeed, I can confirm from looking at the GLFW source that comment in their header seems incorrect. I will send a PR upstream to fix the comment.

slimsag: Sent https://github.com/glfw/glfw/pull/1992

glfw/src/Window.zig
```diff
@@ -205,66 +203,123 @@ pub const Hint = enum(c_int) {
...
-            switch (pointed_type) {
-                .Array => |arr_type| {
-                    if (arr_type.child != u8) {
-                        @compileError("expected pointer to array of u8, got " ++ @typeName(arr_type));
-                    }
-                },
-                else => @compileError("expected pointer to array, got " ++ @typeName(pointed_type)),
+/// Window hints
+pub const Hints = struct {
+    resizable: bool = true,
```
slimsag, line 208:
```suggestion
    // Note: The defaults here are directly from the GLFW source of the glfwDefaultWindowHints function
    resizable: bool = true,
```

slimsag: Adding this note so that others will know how to verify if these values are correct or not and how to update them in the future

glfw/src/Window.zig
```diff
@@ -205,66 +203,124 @@ pub const Hint = enum(c_int) {
...
-            const pointed_type = @typeInfo(pointer_info.child);
-            switch (pointed_type) {
-                .Array => |arr_type| {
-                    if (arr_type.child != u8) {
-                        @compileError("expected pointer to array of u8, got " ++ @typeName(arr_type));
-                    }
-                },
-                else => @compileError("expected pointer to array, got " ++ @typeName(pointed_type)),
+/// Window hints
+pub const Hints = struct {
```
slimsag, line 207: Because these are the only public hints exposed by the API, I think that we need to move all the docstrings from the current `const Hint = enum(c_int) {` into the fields here. Otherwise, once a Zig documentation generator exists (or even if you're just looking at this code for usage) you would not find any docstrings on these fields.

glfw/src/Window.zig
```diff
@@ -205,66 +203,124 @@ pub const Hint = enum(c_int) {
...
+        lose_context_on_reset = c.GLFW_LOSE_CONTEXT_ON_RESET,
+    };
+    
+    pub const ContextReleaseBehavior = enum(c_int) {
+        any_release_behavior = c.GLFW_ANY_RELEASE_BEHAVIOR,
+        release_behavior_flush = c.GLFW_RELEASE_BEHAVIOR_FLUSH,
+        release_behavior_none = c.GLFW_RELEASE_BEHAVIOR_NONE,
+    };
+    
+    pub const OpenGlProfile = enum(c_int) {
```
slimsag, line 290:
```suggestion
    pub const OpenGLProfile = enum(c_int) {
```

glfw/src/Window.zig
```diff
@@ -205,66 +203,124 @@ pub const Hint = enum(c_int) {
...
+    context_release_behavior: ContextReleaseBehavior = .any_release_behavior,
+    
+    /// Note: disables the context creating errors,
+    /// instead turning them into undefined behavior.
+    context_no_error: bool = false,
+    
+    opengl_forward_compat: bool = false,
+    opengl_debug_context: bool = false,
+    
+    opengl_profile: OpenGlProfile = .opengl_any_profile,
```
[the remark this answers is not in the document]
slimsag: And same thing for `ClientApi` and `ContextCreationApi` I think (use `API`)

glfw/src/Window.zig
```diff
@@ -381,7 +437,10 @@ pub inline fn hint(h: Hint, value: anytype) Error!void {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: window_creation, glfw.Window.destroy
-pub inline fn create(width: usize, height: usize, title: [*c]const u8, monitor: ?Monitor, share: ?Window) Error!Window {
+pub inline fn create(width: usize, height: usize, title: [*c]const u8, monitor: ?Monitor, share: ?Window, hints: Hints) Error!Window {
+    try hints.set();
+    try Window.defaultHints();
```
slimsag, line 442: I believe that this is wrong? It would set the user-specified hints, then overwrite them with the defaults?

glfw/src/consts.zig
```diff
@@ -2,30 +2,5 @@
...
-pub const release_behavior_flush = c.GLFW_RELEASE_BEHAVIOR_FLUSH;
-pub const release_behavior_none = c.GLFW_RELEASE_BEHAVIOR_NONE;
-
-/// Possible values for glfw.Window.Hint.context_creation_api hint
-pub const native_context_api = c.GLFW_NATIVE_CONTEXT_API;
-pub const egl_context_api = c.GLFW_EGL_CONTEXT_API;
-pub const osmesa_context_api = c.GLFW_OSMESA_CONTEXT_API;
-
 /// Possible value for various window hints, etc.
 pub const dont_care = c.GLFW_DONT_CARE;
```
slimsag, line 6: Maybe delete this whole file now and just move this constant into `main.zig`, since we already do a `usingnamespace` import of this file: https://github.com/hexops/mach/blob/d5913c3e4961b1d4f2562900b26b5ef2b7280ed6/glfw/src/main.zig#L8

slimsag (approved):
Looks really great, thanks for sending this. Once my comments are addressed and tests pass I'll merge. Feel free to send the test in this PR or a separate one, whichever you prefer.

glfw/src/Window.zig
InKryption: You are correct, that code is wrong. That probably explains why I was getting some strange results prototyping those hint-attribute tests I mentioned.

glfw/src/Window.zig
```diff
@@ -390,6 +448,8 @@ pub inline fn create(width: usize, height: usize, title: [*c]const u8, monitor:
         if (share) |w| w.handle else null,
     );
     try getError();
+    
+    try Window.defaultHints();
```
slimsag, line 452: Ah, I see your intent now is to "cleanup" the hints so-to-speak by changing them back to the default. This makes sense.

However, if window creation fails we would not clean up. Maybe instead `defer` and ignore the error? I think `defer _ = Window.defaultHints();` would do the trick (but I might have the exact syntax wrong)

InKryption: In my original thinking, a `defer` wouldn't have worked, as you can't `try` in a `defer` statement; but looking at it closer, the only possible error for `Window.defaultHints` is `Error.NotInitialized`, which should already have been caught in `Hints.set`, so an error should never be thrown there, a la `catch unreachable`.

glfw/src/Window.zig
InKryption: Here, I'm actually unsure if all of the comments in `Hint` are totally necessary. They about double the length of the struct's declaration, and don't (in my opinion) add that much information. All of them follow the pattern:
```zig
    /// Window {name} hint
    {name},
```
Save for a few, like `center_cursor` and `focus_on_show`, and the framebuffer bits hints, which follow their own pattern.
But if you feel strongly about it, I can add the changes.

slimsag: Makes sense to me! I agree having comments just for the sake of it is bad practice, we should remove all such instances. They probably just ended up there as I didn't notice it. I should've checked the comments were actually useful before asking you to do this, sorry 😛

InKryption: All's good. I'll sort through them for a bit, and grab the ones that are meaningful, and polish them up.

glfw/src/Window.zig
```diff
@@ -68,16 +69,13 @@ pub const InternalUserPointer = struct {
...
 ///
 /// see also: window_hints, glfw.Window.hint, glfw.Window.hintString
-pub inline fn defaultHints() Error!void {
+inline fn defaultHints() Error!void {
     c.glfwDefaultWindowHints();
     try getError();
 }
 
 /// Window hints
 pub const Hint = enum(c_int) {
```
slimsag, line 78: I think this became public again by accident / due to merge conflict

slimsag:
Looks good to me, let me know when you're finished and I'll merge. I will need to squash merge due to non-linear history, unless you want to rebase everything.

glfw/src/Window.zig
InKryption: I've done the sorting; give it a quick look and see if there's anything amiss. If that looks good, then I feel confident about saying this is locked and loaded. I'll look at making a PR for that test/those tests concerning hints/attributes at some point in the near-ish future.

slimsag (approved):
LGTM, thanks a ton again!

InKryption:
Closes #65

PR #72, opened 2021-11-12
glfw: patch more undefined behaviour

silversquirl:
Pressing any given modifier key twice caused ubsan crashes on X11


slimsag:
Interesting find! Is there an upstream GLfW PR for this or is it already fixed in the latest revision upstream?

silversquirl:
Yes, I've PR'd the same fix upstream at glfw/glfw#1989

slimsag (approved):

PR #80, opened 2021-11-17
glfw: reorganize `Window.Attrib` to be in the same style as `Window.Hint`

InKryption:
Compared to `Window.Hint`, where some comments were worth keeping, all comments within `Window.Attribute` appear to be basically redundant. 


InKryption:
Note: moved `focused` down under `decorated`, the same as done in `Window.Hint`, which mirrors the documentation for `Window.Hint`, to make it easier to cross-reference between here and canonical GLFW documentation.

slimsag (approved):
Nice, thank you!

PR #81, opened 2021-11-18
glfw: window hint default values parity test with attributes

InKryption:
Tests pass on my machine right now, but also unsure whether this is as comprehensive as it could be, hence the mess of comments. Any feedback or insights would be greatly appreciated. 

### DIFF
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -3221,3 +3221,104 @@ test "setScrollCallback" {
         }
     }).callback);
 }
+
+fn failedToCreateWindow(err: Error) void {
+    // return without fail, because most of our CI environments are headless / we cannot open
+    // windows on them.
+    std.debug.print("note: failed to create window: {}\n", .{err});
+    return;
+}
+
+test "hint-attribute default value parity" {
+    try glfw.init(.{});
+    defer glfw.terminate();
+    const window_a = @as(Error!Window, handle: {
+        // manually create window, to avoid call to `Window.Hints.set` in `Window.create`, which
+        // would overwrite the default hint values set by GLFW itself.
+        const handle = c.glfwCreateWindow(640, 480, "Hello, Zig!", null, null);
+        getError() catch |err| break :handle err;
+        break :handle Window.from(handle.?) catch |err| break :handle err;
+    }) catch |err| return failedToCreateWindow(err);
+    defer window_a.destroy();
+
+    const window_b = Window.create(640, 480, "Hello, Zig!", null, null, .{}) catch |err| return failedToCreateWindow(err);
+    defer window_b.destroy();
+
+    inline for (comptime std.enums.values(Window.Hint)) |hint_tag| {
+        if (@hasField(Window.Attrib, @tagName(hint_tag))) {
+            const attrib_tag = @field(Window.Attrib, @tagName(hint_tag));
+            switch (attrib_tag) {
+                .resizable,
+                .visible,
+                .decorated,
+                .auto_iconify,
+                .floating,
+                .maximized,
+                .transparent_framebuffer,
+                .focus_on_show,
+                .client_api,
+                .context_creation_api,
+                .context_version_major,
+                .context_version_minor,
+                .context_robustness,
+                .context_release_behavior,
+                .opengl_forward_compat,
+                .opengl_debug_context,
+                .opengl_profile,
+                => {
+                    const expected = window_a.getAttrib(attrib_tag) catch |err| {
+                        std.debug.print("Failed to get attribute '{}' value from window_a with error '{}'.\n", .{ attrib_tag, err });
+                        return;
+                    };
+                    const actual = window_b.getAttrib(attrib_tag) catch |err| {
+                        std.debug.print("Failed to get attribute '{}' value from window_b with error '{}'.\n", .{ attrib_tag, err });
+                        return;
+                    };
+
+                    testing.expectEqual(expected, actual) catch |err| {
+                        std.debug.print("On attribute '{}'.\n", .{hint_tag});
+                        return err;
+                    };
+                },
+
+                // This attribute is based on a check for which window is currently in focus,
+                // and the default value, as of writing this comment, is 'true', which means
+                // that first window_a takes focus, and then window_b takes focus, meaning
+                // that we can't actually test for the default value.
+                .focused => continue,
+
+                // TODO: consider this one differently, as we've chosen a different default value
+                .context_no_error => continue,
+
+                .iconified,
+                .hovered,
+                .context_revision,
+                => unreachable,
+            }
+        }
+        // TODO: consider hint values that can't be retrieved via attributes:
+        // center_cursor
+        // scale_to_monitor
+        // red_bits
+        // green_bits
+        // blue_bits
+        // alpha_bits
+        // depth_bits
+        // stencil_bits
+        // accum_red_bits
+        // accum_green_bits
+        // accum_blue_bits
+        // accum_alpha_bits
+        // aux_buffers
+        // samples
+        // refresh_rate
+        // stereo
+        // srgb_capable
+        // doublebuffer
+
+        // platform specific, and thus not considered:
+        // cocoa_retina_framebuffer
+        // cocoa_frame_name
+        // cocoa_graphics_switching
+    }
+}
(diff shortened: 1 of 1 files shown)


slimsag:
Looks like it passed on Windows, Linux, and Mac (x86_64 + aarch64)

slimsag:
* Linux: confirmed this passed (did not log an error) in CI
* Windows: I haven't tested, CI for Windows is currently headless so I'll need to test there manually
* Mac: Haven't tested, our x86_64 CI is headless

slimsag:
Can you clarify which OS your machine is?

glfw/src/Window.zig
```diff
@@ -3221,3 +3221,104 @@ test "setScrollCallback" {
...
+test "hint-attribute default value parity" {
+    try glfw.init(.{});
+    defer glfw.terminate();
+    const window_a = @as(Error!Window, handle: {
+        // manually create window, to avoid call to `Window.Hints.set` in `Window.create`, which
+        // would overwrite the default hint values set by GLFW itself.
+        const handle = c.glfwCreateWindow(640, 480, "Hello, Zig!", null, null);
+        getError() catch |err| break :handle err;
+        break :handle Window.from(handle.?) catch |err| break :handle err;
+    }) catch |err| return failedToCreateWindow(err);
```
slimsag, line 3241: thought: maybe using a global var `testing_set_no_hints` and `comptime if` inside `Window.create` to disable the hinting calls would be simpler than all of this logic?

glfw/src/Window.zig
```diff
@@ -3221,3 +3221,104 @@ test "setScrollCallback" {
...
+    defer window_a.destroy();
+
+    const window_b = Window.create(640, 480, "Hello, Zig!", null, null, .{}) catch |err| return failedToCreateWindow(err);
```
slimsag, line 3244: All else equal, I'd prefer to keep tests a bit isolated so maybe ditch the `failedToCreateWindow(err)` in this case

glfw/src/Window.zig
```diff
@@ -3221,3 +3221,104 @@ test "setScrollCallback" {
...
+                    };
+                },
+
+                // This attribute is based on a check for which window is currently in focus,
+                // and the default value, as of writing this comment, is 'true', which means
+                // that first window_a takes focus, and then window_b takes focus, meaning
+                // that we can't actually test for the default value.
+                .focused => continue,
+
+                // TODO: consider this one differently, as we've chosen a different default value
```
slimsag, line 3290: What do you mean by this? "as we've chosen a different default value"?

glfw/src/Window.zig
```diff
@@ -3221,3 +3221,104 @@ test "setScrollCallback" {
...
+        // samples
+        // refresh_rate
+        // stereo
+        // srgb_capable
+        // doublebuffer
+
+        // platform specific, and thus not considered:
+        // cocoa_retina_framebuffer
+        // cocoa_frame_name
+        // cocoa_graphics_switching
```
slimsag, line 3322: not worried about these

glfw/src/Window.zig
```diff
@@ -3221,3 +3221,104 @@ test "setScrollCallback" {
...
+        // accum_red_bits
+        // accum_green_bits
+        // accum_blue_bits
+        // accum_alpha_bits
+        // aux_buffers
+        // samples
+        // refresh_rate
+        // stereo
+        // srgb_capable
+        // doublebuffer
```
slimsag, line 3317: in recent GLFW versions, doublebuffer can also only be queried not set (which is super reasonable)

I'd say it's fine if we do not test all of these.

slimsag:
I left some thoughts inline, generally this looks great to me and it's OK if we don't have perfect coverage of every possible hint IMO. Just having something in place here for general sanity works for me.

InKryption:
@slimsag Sorry, it's been a rough couple of days for me. I'll try and get to your comments. And also, my primary OS is currently Windows 10, so tests passed there. Should have clarified that.

glfw/src/Window.zig
InKryption: That's fair. I made it because I was running into some weird issues with block return and 'catching' errors from it, but I eventually figured it out, just didn't think to remove it.

glfw/src/Window.zig
InKryption: Are you implying we use a global `comptime var`? Because I don't think that's possible. Or do you mean something like `const testing_set_no_hints = std.meta.globalOption("glfw_testing_set_no_hints") orelse false`? I don't know if that would work, since I don't think (?) tests are treated as "root", though I'll have to check that.

glfw/src/Window.zig
InKryption: In the discussion [here](https://github.com/hexops/mach/pull/71#discussion_r748416222) in my window hints rework pr, me and @silversquirl came to the accord that, seeing as GLFW doesn't specify a default value for `context_no_error` in their docs in [this list](https://www.glfw.org/docs/latest/window_guide.html#window_hints_values), it would be most sane to default it to false (seeing as, in zig, we almost always want to handle errors with error codes, and avoid UB).

Though, thinking about it now, and looking at the source code, it actually is defaulted to false (albeit not directly, it's done through a call to `memset`), so there's no actual difference in behavior here. Though, maybe this calls for a pr upstream to add its default value to the aforementioned list in their docs? Then I could amend this "patch" in this pr.

glfw/src/Window.zig
InKryption: A different solution occurred to me. Make a normal global variable, but with this declaration:
```zig
var testing_ignore_hints_struct = if (@import("builtin").is_test) false else @as(void, {});
```
Which would change the code inside `create` to:
```zig
const ignore_hints_struct = if (comptime @import("builtin").is_test) testing_ignore_hints_struct else false;
if (!ignore_hints_struct) {
    try hints.set();
    defer defaultHints() catch unreachable; // this should be unreachable, being that this should be caught in the previous call to `Hints.set`.
}
```

InKryption: I've committed the changes. I'll leave this as unresolved in case you have anything you want to clarify or suggest.

glfw/src/Window.zig
slimsag: Ahh okay, I see. I was aware of the fact that it was implicitly set to false in GLFW through `memset` (I double checked the defaults when I reviewed, which was why I was surprised to see this TODO)

In that case, I am happy with this as-is and think we can just remove the TODO mention here.

slimsag:
@InKryption thanks a ton! And sorry to hear that :) FWIW your PRs have made my day for the past week or so, I really appreciate what you're doing, If I can buy you a coffee or two, let me know.

glfw/src/Window.zig
```diff
@@ -3221,3 +3226,105 @@ test "setScrollCallback" {
...
+                // TODO: consider this one differently, as we've chosen a different default value
+                .context_no_error => continue,
+
+                .iconified,
+                .hovered,
+                .context_revision,
+                => unreachable,
+            }
+        }
+        // TODO: consider hint values that can't be retrieved via attributes:
```
slimsag, line 3304:
```suggestion
        // Future: we could consider hint values that can't be retrieved via attributes:
```

glfw/src/Window.zig
```diff
@@ -3221,3 +3226,105 @@ test "setScrollCallback" {
...
+                    };
+                },
+
+                // This attribute is based on a check for which window is currently in focus,
+                // and the default value, as of writing this comment, is 'true', which means
+                // that first window_a takes focus, and then window_b takes focus, meaning
+                // that we can't actually test for the default value.
+                .focused => continue,
+
+                // TODO: consider this one differently, as we've chosen a different default value
```
slimsag, line 3295:
```suggestion
```

slimsag:
Pretty much LGTM!

glfw/src/Window.zig
slimsag: Smart approach - I wouldn't have thought of that. Nice work.

slimsag (approved):

InKryption:
@slimsag Appreciate the thought; and no worries, I really like what you have here, and thought I ought to try and help make it better, both for my own purposes, and as an added bonus, for others. As well, seemed like a friendly enough place to try making some more significant contributions for the first time.

PR #98, opened 2021-11-23
glfw: Eliminate `InvalidEnum`

InKryption:
An initial effort to eliminate the need for `InvalidEnum`; there are a couple of places preventing straight removal of the error, so I'll have a look at the GLFW source to see if they can be omitted safely. 

### DIFF
--- a/glfw/src/Joystick.zig
+++ b/glfw/src/Joystick.zig
@@ -91,7 +91,6 @@ pub inline fn present(self: Joystick) Error!bool {
     internal_debug.assertInitialized();
     const is_present = c.glfwJoystickPresent(@enumToInt(self.jid));
     getError() catch |err| return switch (err) {
-        Error.InvalidEnum => unreachable, // intentionally invalid enum value
         Error.PlatformError => err,
         else => unreachable,
     };
@@ -123,7 +122,6 @@ pub inline fn getAxes(self: Joystick) Error!?[]const f32 {
     var count: c_int = undefined;
     const axes = c.glfwGetJoystickAxes(@enumToInt(self.jid), &count);
     getError() catch |err| return switch (err) {
-        Error.InvalidEnum => unreachable, // intentionally invalid enum value
         Error.PlatformError => err,
         else => unreachable,
     };
@@ -160,7 +158,6 @@ pub inline fn getButtons(self: Joystick) Error!?[]const u8 {
     var count: c_int = undefined;
     const buttons = c.glfwGetJoystickButtons(@enumToInt(self.jid), &count);
     getError() catch |err| return switch (err) {
-        Error.InvalidEnum => unreachable, // intentionally invalid enum value
         Error.PlatformError => err,
         else => unreachable,
     };
@@ -213,7 +210,6 @@ pub inline fn getHats(self: Joystick) Error!?[]const Hat {
     var count: c_int = undefined;
     const hats = c.glfwGetJoystickHats(@enumToInt(self.jid), &count);
     getError() catch |err| return switch (err) {
-        Error.InvalidEnum => unreachable, // intentionally invalid enum value
         Error.PlatformError => err,
         else => unreachable,
     };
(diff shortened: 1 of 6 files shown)


InKryption:
Looking again, I now remember because I left myself a handy comment, 'setInputMode' is the main culprit here, whose parameter 'value' has valid range of values, which is dependent on the 'mode' parameter at runtime. I see a few solutions here:
1. We make 'mode' comptime
2. We assert that 'value' has the correct type based on the value of 'mode' at runtime
3. Me privatize 'setInputMode', and refer users to its variants

I would opt for solution 1 or 2.

slimsag:
Looks great - I think I'd opt for solution 1

InKryption:
Sounds good; guess it makes sense, seeing as I can't actually think of a use-case for determining 'mode' at runtime at the call-site.

slimsag:
Worst case scenario someone could create their own enum to do so and call the right fns with the right params.

glfw/src/Joystick.zig
```diff
@@ -424,10 +418,7 @@ pub inline fn updateGamepadMappings(gamepad_mappings: [*:0]const u8) Error!void
 pub inline fn isGamepad(self: Joystick) bool {
     internal_debug.assertInitialized();
     const is_gamepad = c.glfwJoystickIsGamepad(@enumToInt(self.jid));
-    getError() catch |err| return switch (err) {
-        Error.InvalidEnum => unreachable, // intentionally invalid enum value
-        else => unreachable,
-    };
+    getError() catch unreachable;
```
slimsag, line 421: Thought for the future: we should change all such locations like this to a @panic maybe so we know what error occurred (e.g. if we see this in the wild, what error caused it? which one is not being handled?)

InKryption: That's a great idea; though I think that deserves its own PR, dedicated to just replacing all the 'catch unreachable's.

glfw/src/Window.zig
```diff
@@ -1610,6 +1600,16 @@ pub const InputMode = enum(c_int) {
     sticky_mouse_buttons = c.GLFW_STICKY_MOUSE_BUTTONS,
     lock_key_mods = c.GLFW_LOCK_KEY_MODS,
     raw_mouse_motion = c.GLFW_RAW_MOUSE_MOTION,
+    
+    pub fn Value(comptime input_mode: InputMode) type {
```
slimsag, line 1604:
```suggestion
    /// Tells which value type is required for the given input mode when invoking `setInputMode`.
    /// Generally, you should use one of the `setInputMode` variants which does not take a `mode`
    /// parameter instead.
    pub fn Value(comptime input_mode: InputMode) type {
```

slimsag: The more I look at this, the restriction we impose on having `input_mode` be `comptime`, and the fact that we do have the typed variants... I wonder: should `setInputMode` actually be public?

The original thinking was that at least it allowed for someone to port some existing GLFW code over that maybe stores the input mode constant at runtime, but now that we're eliminating that possibility and it's only a comptime thing.. hmm.

slimsag: Maybe we should make it private, and if anyone complains about not having the ability to do a runtime `setInputMode` / having to write their own code to do that, we re-evaluate?

InKryption: Well, the other possibility here is my second proposition, "We assert that 'value' has the correct type based on the value of 'mode' at runtime", which has the same effect of allowing us to eliminate 'InvalidEnum', and the added clause that it would allow easier porting of code like you say.

slimsag: Yes, true, I guess that does make more sense. Sorry for not catching that earlier, I'm quite under the weather atm so my mind is a bit foggy.

InKryption: Aye, no worries mate, I know the feeling. That's why we have these review discussions at the end of the day; just gotta get on the same page.

slimsag (approved):
LGTM, I'll merge now if you agree this is good to go

InKryption:
Pardon, fell asleep right after you sent that it seems. Time zones are pretty off.

slimsag (approved):

PR #104, opened 2021-11-26
glfw: Eliminate `Error.InvalidValue`

### LINKED ISSUE #101: glfw: Eliminate `Error.InvalidValue`
Before any work is done to progress #96, I think it wise to first mop up the errors we can potentially get rid of; that's already done for the now defunct `Error.InvalidEnum`, but that remains to be done for `Error.InvalidValue`.

This issue can serve to track discussion on that effort, but the main reason I'm opening this is because I've hit a roadblock, and would like to hear some feedback: how to deal with `Joystick.updateGamepadMappings`?
The GLFW call does some parsing of the passed string, and can set 'InvalidValue' if it encounters anythign unexpected, which makes preventing this error condition quite a bit more complex.

One option is to try and mimic GLFW's parsing to pre-emptively assert that the format is wrong, and perhaps deliver a useful error message to inform the caller of what's gone wrong; the other option is to simply leave the `Error.InvalidValue` code for this function, and allow it to be unique to it, and any other functions that may do parsing that we can't pre-empt in a maintainable fashion, which, while it doesn't allow us to eliminate the error altogether as hoped, would still be more or less made okay by error denormalization.

InKryption:
Closes #101.

As will be noted, the actual error hasn't been deleted, but measures have been taken such that it will never be encountered as an error by the user, save for in the case of `Joystick.updateGamepadMappings`, which will be changed to have its own error `ParseError` later on, when errors are denormalized. 

### DIFF
--- a/glfw/src/time.zig
+++ b/glfw/src/time.zig
@@ -53,12 +53,18 @@ pub inline fn getTime() f64 {
 /// base time is not atomic, so it needs to be externally synchronized with calls to glfw.getTime.
 ///
 /// see also: time
-pub inline fn setTime(time: f64) Error!void {
+pub inline fn setTime(time: f64) void {
     internal_debug.assertInitialized();
+
+    std.debug.assert(!std.math.isNan(time));
+    std.debug.assert(time >= 0);
+    // TODO: Look into why GLFW uses this hardcoded float literal as the maximum valid value for 'time'.
+    // Maybe propose upstream to name this constant. 
+    std.debug.assert(time <= 18446744073.0);
+
     c.glfwSetTime(time);
     getError() catch |err| return switch (err) {
-        // TODO: Consider whether to use GLFW error handling, or assert that 'time' is a valid value
-        Error.InvalidValue => err,
+        Error.InvalidValue => unreachable, // we assert that 'time' is a valid value, so this should be impossible
         else => unreachable,
     };
 }
@@ -109,7 +115,7 @@ test "setTime" {
     try glfw.init(.{});
     defer glfw.terminate();
 
-    _ = try glfw.setTime(1234);
+    _ = glfw.setTime(1234);
 }
 
 test "getTimerValue" {
--- a/glfw/src/Joystick.zig
+++ b/glfw/src/Joystick.zig
@@ -395,7 +395,11 @@ pub inline fn updateGamepadMappings(gamepad_mappings: [*:0]const u8) Error!void 
     internal_debug.assertInitialized();
     _ = c.glfwUpdateGamepadMappings(gamepad_mappings);
     getError() catch |err| return switch (err) {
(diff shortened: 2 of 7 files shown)


glfw/src/time.zig
```diff
@@ -53,12 +53,18 @@ pub inline fn getTime() f64 {
...
 /// see also: time
-pub inline fn setTime(time: f64) Error!void {
+pub inline fn setTime(time: f64) void {
     internal_debug.assertInitialized();
+
+    std.debug.assert(!std.math.isNan(time));
+    std.debug.assert(time >= 0);
+    // TODO: Look into why GLFW uses this hardcoded float literal as the maximum valid value for 'time'.
+    // Maybe propose upstream to name this constant. 
+    std.debug.assert(time <= 18446744073.0);
```
slimsag, line 63: Very odd, no clue myself.

slimsag (approved):
Looks really great to me so far.

slimsag:
(I notice it's still a draft, so I'll hold off on merging)

glfw/src/time.zig
InKryption: Too bad, was hoping you maybe had an inkling.

InKryption:
This is a lot of "stub" bits to lay the groundwork for further work; if you don't have any notes, then this is ready to merge on my part.

glfw/src/time.zig
slimsag: I told my brother @Andoryuuta about this, he found the answer for us :)

The initial digits match up with 2^64, which is true:

```
18446744073.0                # GLFW constant
18446744073709551616 # 2^64
```

But why don't the remainder? Well, 1 second = 1000000000 nanoseconds, so if we multiply the constant times nanoseconds we get:

```
18446744073.0 * 1000000000  == 18446744073000000000
```

Now if we compare we see:

```
18446744073000000000 # GLFW constant in nanoseconds
18446744073709551616  # 2^64
```

And so it becomes clear: `18446744073000000000` (GLFW constant in nanoseconds) is the largest second representable by a u64 type with nanosecond precision (if we add another `1000000000`, we'd overflow)

InKryption: Huh, fancy that. Well we should probably add a comment about that in case anyone else asks a similar question in the future.

slimsag: Probably it was easier to hard-code that way due to C always being confusing :) but if one can rely on division and integer truncation, this would be a much clearer way to write it I think (Python):

```
((2**64) / (1000*1000*1000)) # largest number of seconds that can fit in uint64 with nanosecond precision
```

I'll leave the Zig port to you :)

InKryption: Went ahead and made a quick PR for it.

PR #105, opened 2021-11-27
glfw: un-hardcode constant

InKryption:
Make the constant extracted from GLFW source into `time.setTime`, into a better-documented set of comptime instructions

### DIFF
--- a/glfw/src/time.zig
+++ b/glfw/src/time.zig
@@ -58,9 +58,11 @@ pub inline fn setTime(time: f64) void {
 
     std.debug.assert(!std.math.isNan(time));
     std.debug.assert(time >= 0);
-    // TODO: Look into why GLFW uses this hardcoded float literal as the maximum valid value for 'time'.
-    // Maybe propose upstream to name this constant. 
-    std.debug.assert(time <= 18446744073.0);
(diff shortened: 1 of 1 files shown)


glfw/src/time.zig
```diff
@@ -58,9 +58,11 @@ pub inline fn setTime(time: f64) void {
 
     std.debug.assert(!std.math.isNan(time));
     std.debug.assert(time >= 0);
-    // TODO: Look into why GLFW uses this hardcoded float literal as the maximum valid value for 'time'.
-    // Maybe propose upstream to name this constant. 
-    std.debug.assert(time <= 18446744073.0);
+    // assert time is lteq to largest number of seconds representable by u64 with nanosecond precision
+    std.debug.assert(time <= max_time: {
+        const @"2^64" = std.math.maxInt(u64);
```
slimsag, line 63: Ah I forgot Zig doesn't have an integer power operator, we could also use `2 << 63` here.

slimsag (approved):
Great!

glfw/src/time.zig
InKryption: That's true, but I thought I'd try and use the type "u64" directly in the code to indicate it is involved, thus `maxInt`.

PR #111, opened 2021-11-29
glfw: Fix pointer cast

InKryption:
The GLFW implementation takes a constant pointer to a `VkAllocationCallbacks` struct; casting it to a mutable pointer equivalent here doesn't break anything, but does prevent passing a valid const pointer, which is often what one should prefer to do.
As well, the `@alignOf` builtin takes the alignment expected of the type directly, so `@alignOf(*T)` returns the alignment of `*T`, not `T`, so that has also been corrected.


slimsag (approved):
Nice catch, I appreciate this a lot!

PR #115, opened 2021-12-05
glfw: Denormalize errors

### LINKED ISSUE #96: glfw: Denormalize Errors
In the original draft of #92, I added an off-the-cuff change to de-normalize error sets. As suggested by @slimsag, I'm opening an issue to discuss this change (and whether it's beneficial enough to consider adding).

Currently, we have normalized errors, adhering to the general possibility imposed by GLFW of (almost) any function returning (or rather "setting") any of the errors specified by the library. But it is important to note that GLFW does obviously specify, for each function, its "possible" error set, and according to the maintainer of GLFW, this documentation is "generally accurate" (paraphrased).

A consideration here is ABI stability: with a normalized error set, that's (almost universally) a guarantee, whilst with denormalized error sets, that's essentially impossible unless we were to freeze the GLFW version.

The other consideration here is how "comfortable" the API is: while a comment describing which errors you might encounter is alright, it is a departure from the error handling philosophy of Zig, and much more akin to C (which is what GLFW aims to cater to). Denormalized errors would slightly increase maintenance, but it would vastly improve readability, given that the errors the user would have to handle would be in the function's signature, and makes it such that a compilation error will be emitted if the user tries to catch an error which the function can't emit. 

As well, another positive of denormalized errors would be behaviour-documentation parity. If GLFW changes the behaviour of one of their functions, and adds a possible error to a function, then we'll `catch unreachable` on that prong, and be able to make a PR upstream to update their docs. Net positive. 

I am in favour of denormalized errors.

InKryption:
Tests pass on my machine, and everything seems fine; added a TODO note concerning what looks like a memory leak in 'Window.zig'. If there's discussion to be had, I'll turn this into a draft, but otherwise I feel like it's ready to merge. 

Also brought back errors that had been deleted in the 'Error' error set, and wrote out some unreachable paths that were previously implicitly unreachable, being that the Error error set is no longer forwarded uniformly; adds some documentative value. 

Closes #96.

### DIFF
--- a/glfw/src/Cursor.zig
+++ b/glfw/src/Cursor.zig
@@ -60,12 +60,13 @@ pub const Shape = enum(isize) {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: cursor_object, glfw.Cursor.destroy, glfw.Cursor.createStandard
-pub inline fn create(image: Image, xhot: isize, yhot: isize) Error!Cursor {
+pub inline fn create(image: Image, xhot: isize, yhot: isize) error{PlatformError}!Cursor {
     internal_debug.assertInitialized();
     const img = image.toC();
     const cursor = c.glfwCreateCursor(&img, @intCast(c_int, xhot), @intCast(c_int, yhot));
     getError() catch |err| return switch (err) {
-        Error.PlatformError => err,
+        Error.NotInitialized => unreachable,
+        Error.PlatformError => @errSetCast(error{PlatformError}, err),
         else => unreachable,
     };
     return Cursor{ .ptr = cursor.? };
@@ -80,11 +81,13 @@ pub inline fn create(image: Image, xhot: isize, yhot: isize) Error!Cursor {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: cursor_object, glfwCreateCursor
-pub inline fn createStandard(shape: Shape) Error!Cursor {
+pub inline fn createStandard(shape: Shape) error{PlatformError}!Cursor {
     internal_debug.assertInitialized();
     const cursor = c.glfwCreateStandardCursor(@intCast(c_int, @enumToInt(shape)));
     getError() catch |err| return switch (err) {
-        Error.PlatformError => err,
+        Error.NotInitialized => unreachable,
+        Error.InvalidEnum => unreachable,
+        Error.PlatformError => @errSetCast(error{PlatformError}, err),
         else => unreachable,
     };
     return Cursor{ .ptr = cursor.? };
@@ -109,7 +112,7 @@ pub inline fn destroy(self: Cursor) void {
     internal_debug.assertInitialized();
     c.glfwDestroyCursor(self.ptr);
     getError() catch |err| return switch (err) {
-        Error.PlatformError => std.log.debug("{}: was unable to destroy Cursor.\n", .{err}),
+        Error.PlatformError => std.log.err("{}: was unable to destroy Cursor.\n", .{err}),
         else => unreachable,
     };
 }
--- a/glfw/src/Cursor.zig
+++ b/glfw/src/Cursor.zig
@@ -112,7 +112,7 @@ pub inline fn destroy(self: Cursor) void {
     internal_debug.assertInitialized();
     c.glfwDestroyCursor(self.ptr);
     getError() catch |err| return switch (err) {
-        Error.PlatformError => std.log.err("{}: was unable to destroy Cursor.\n", .{err}),
+        Error.PlatformError => std.log.err("mach/glfw: unable to destroy Cursor: {}\n", .{err}),
         else => unreachable,
     };
 }
--- a/glfw/src/Joystick.zig
+++ b/glfw/src/Joystick.zig
@@ -87,11 +87,13 @@ const GamepadState = extern struct {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: joystick
-pub inline fn present(self: Joystick) Error!bool {
+pub inline fn present(self: Joystick) error{PlatformError}!bool {
     internal_debug.assertInitialized();
     const is_present = c.glfwJoystickPresent(@enumToInt(self.jid));
     getError() catch |err| return switch (err) {
-        Error.PlatformError => err,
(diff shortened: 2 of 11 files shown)


InKryption:
Also, closes #96

glfw/src/Joystick.zig
```diff
@@ -438,17 +464,21 @@ pub inline fn isGamepad(self: Joystick) bool {
 /// @return The UTF-8 encoded name of the gamepad, or null if the joystick is not present or does
 /// not have a mapping.
 ///
+/// Possible errors include glfw.Error.NotInitialized and glfw.Error.InvalidEnum.
+///
```
slimsag, line 468: Ah, just to make sure I understand this addition, we're adding it to keep in sync with GLFW documentation since every other function has these - but the function can't actually return these.

Is `InvalidEnum` as is documented here correct? Should it be `InvalidValue` instead?

glfw/src/Window.zig
```diff
@@ -1361,9 +1416,15 @@ pub inline fn setPosCallback(self: Window, callback: ?fn (window: Window, xpos:
...
     _ = c.glfwSetWindowPosCallback(self.handle, if (callback != null) setPosCallbackWrapper else null);
-    getError() catch unreachable; // Only error 'GLFW_NOT_INITIALIZED' is impossible
+    getError() catch |err| return switch (err) {
+        Error.NotInitialized => unreachable,
+        else => unreachable,
+    };
 }
 
+// TODO: don't the calls to `from` in these `set*CallbackWrapper` functions cause a leak?
+// is this what's being reported by valgrind in [#60](https://github.com/hexops/mach/issues/60)?
```
slimsag, line 1426: Hmm, `from` is only going to allocate a user pointer for the window if one doesn't already exist, though.

That is, as long as the callback wrapper is invoked with the `handle` of a window created by mach/glfw (and not some other GLFW window created outside mach/glfw), then inside of `from` `glfwGetWindowUserPointer` should return a non-null pointer and no leak should happen.

also note that inside of `Window.destroy` we only destroy the user pointer object _after_ we've destroyed the window, so no calls to this callback should be issued afterwards either. https://github.com/hexops/mach/blob/233fc53a17d16ab4901bda9f7bd332663e750c75/glfw/src/Window.zig#L448-L450

So I don't think any leak is possible here, but if you think otherwise I am super keen to hear about it!

glfw/src/Window.zig
```diff
@@ -1434,33 +1501,34 @@ fn setRefreshCallbackWrapper(handle: ?*c.GLFWwindow) callconv(.C) void {
     internal.setRefreshCallback.?(window);
 }
 
-// TODO: Something's off about this comment?
```
slimsag, line 1437: thanks for fixing this, not sure how I missed this one

glfw/src/errors.zig
```diff
@@ -4,13 +4,25 @@ const c = @import("c.zig").c;
 
 /// Errors that GLFW can produce.
 pub const Error = error{
+    /// GLFW has not been initialized.
+    ///
+    /// This occurs if a GLFW function was called that must not be called unless the library is
+    /// initialized.
+    NotInitialized,
```
slimsag, line 11: Did you intend to bring this back? I thought that we'd eliminated the need for NotInitialized and InvalidEnum errors

slimsag: Ahh, is it just so that `convertError` can return the error so we can `unreachable` it later inside the function for more explicit handling? that makes sense

glfw/src/main.zig
```diff
@@ -89,6 +92,7 @@ pub inline fn init(hints: InitHints) Error!void {
 ///
 /// Possible errors include glfw.Error.PlatformError.
 ///
+// TODO: Should this remark be removed? Or should we allow this function to be called before init?
 /// remark: This function may be called before glfw.init.
```
slimsag, line 96: Good question.. I would say that in general someone calling terminate without a prior init is likely to be a bug in their code, so I'd err on the side of removing this remark instead.

glfw/src/main.zig
```diff
@@ -101,6 +105,10 @@ pub inline fn terminate() void {
     internal_debug.assertInitialized();
     internal_debug.toggleInitialized();
     c.glfwTerminate();
+    getError() catch |err| return switch (err) {
+        Error.PlatformError => std.log.err("Failed to terminate GLFW: {}", .{err}),
```
slimsag, line 109:
```suggestion
        Error.PlatformError => std.log.err("mach/glfw: Failed to terminate GLFW: {}", .{err}),
```

glfw/src/Cursor.zig
```diff
@@ -109,7 +112,7 @@ pub inline fn destroy(self: Cursor) void {
     internal_debug.assertInitialized();
     c.glfwDestroyCursor(self.ptr);
     getError() catch |err| return switch (err) {
-        Error.PlatformError => std.log.debug("{}: was unable to destroy Cursor.\n", .{err}),
+        Error.PlatformError => std.log.err("{}: was unable to destroy Cursor.\n", .{err}),
```
slimsag, line 115:
```suggestion
        Error.PlatformError => std.log.err("mach/glfw: unable to destroy Cursor: {}\n", .{err}),
```

glfw/src/errors.zig
slimsag: Ah, yep, I should've read PR description first - my bad.

slimsag (approved):
Left some things we can discuss, but LGTM and merging.

glfw/src/Window.zig
slimsag: Maybe we can/should use GeneralPurposeAllocator here to verify we're not leaking

glfw/src/Joystick.zig
InKryption: Yes, that's exactly right; and yes, it is `InvalidEnum`, as I mentioned in https://github.com/glfw/glfw/pull/2007; there's no calls to set `InvalidValue` in any of the top level source or in the possible implementation functions, so that's the set.

glfw/src/Window.zig
InKryption: You're right, there is no leak, not directly observable from what I've tried at least; false alarm on my part.

PR #120, opened 2021-12-10
glfw: replace usage of glfw.dont_care as sentinel value

### LINKED ISSUE #119: glfw: 'dont_care' value limitation in `Window.setSizeLimits`
Currently, the function `Window.setSizeLimits` takes as parameters the window handle, and two `Window.Size`s named 'min' and 'max'. `Window.Size` is a struct with two fields 'width' and 'height', each with type `usize`; a limitation I noticed with this setup is that, unlike what is said in the doc-comment of the function in question, there is no way to pass `glfw.dont_care` as a value to indicate "I don't care what the min/max value of this dimension is relative to the others", because GLFW defines `glfw.dont_care` (`GLFW_DONT_CARE`) as a negative value, which can't be represented by a `usize`.

Possible approaches that come to mind:
 * Make an alternative 'SizeSigned' struct with signed integer 'width' and 'height' fields, which would then allow directly expressing the same usage of the GLFW API as plain old GLFW.
 * Make an alternative 'SizeOptional' struct with optional integer 'width' and 'height' fields, which could then be set to `null` to express the idea of `glfw.dont_care` (in the implementation it would be something like `min.width orelse glfw.dont_care` perhaps) in the Zig style.

I'm personally in favor of the second solution, but can see the value in the first.

It would also probably be a good idea to take this in consideration for the rest of the codebase, wherever else the concept of `glfw.dont_care` applies.

InKryption:
Change some parameters in `glfw.Window` functions to accept null in place of `glfw.dont_care` as a sentinel value.

Closes #119.

This is a quick change, but there are other places where `glfw.dont_care` is an accepted value to ask for "default behavior", such as in hints; should we also change those to be null-able, or leave them as is?

### DIFF
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -704,6 +704,8 @@ pub inline fn setSize(self: Window, size: Size) error{PlatformError}!void {
     };
 }
 
+/// A size with option width/height, used to represent e.g. constraints on a windows size while
+/// allowing specific axis to be unconstrained (null) if desired.
 pub const SizeOptional = struct {
     width: ?usize,
     height: ?usize,
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -1424,9 +1424,6 @@ pub inline fn setPosCallback(self: Window, callback: ?fn (window: Window, xpos: 
     };
 }
 
-// TODO: don't the calls to `from` in these `set*CallbackWrapper` functions cause a leak?
-// is this what's being reported by valgrind in [#60](https://github.com/hexops/mach/issues/60)?
-
 fn setSizeCallbackWrapper(handle: ?*c.GLFWwindow, width: c_int, height: c_int) callconv(.C) void {
     internal_debug.assertInitialized();
     const window = from(handle.?) catch unreachable;
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -272,7 +272,7 @@ pub const Hints = struct {
             const hint_value = @field(hints, field_name);
             switch (@TypeOf(hint_value)) {
                 bool => c.glfwWindowHint(hint_tag, @boolToInt(hint_value)),
-                ?PositiveCInt => c.glfwWindowHint(hint_tag, if (hint_value) |refresh_rate| refresh_rate else glfw.dont_care),
+                ?PositiveCInt => c.glfwWindowHint(hint_tag, if (hint_value) |unwrapped| unwrapped else glfw.dont_care),
                 c_int => c.glfwWindowHint(hint_tag, hint_value),
 
                 ClientAPI,
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -189,7 +189,7 @@ pub const Hints = struct {
     samples: c_int = 0,
 
     /// Monitor refresh rate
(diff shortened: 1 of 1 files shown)


glfw/src/Window.zig
```diff
@@ -704,6 +704,11 @@ pub inline fn setSize(self: Window, size: Size) error{PlatformError}!void {
     };
 }
 
+pub const SizeOptional = struct {
```
slimsag, line 707:
```suggestion
/// A size with option width/height, used to represent e.g. constraints on a windows size while
/// allowing specific axis to be unconstrained (null) if desired.
pub const SizeOptional = struct {
```

glfw/src/Window.zig
```diff
@@ -1422,6 +1424,9 @@ pub inline fn setPosCallback(self: Window, callback: ?fn (window: Window, xpos:
     };
 }
 
+// TODO: don't the calls to `from` in these `set*CallbackWrapper` functions cause a leak?
+// is this what's being reported by valgrind in [#60](https://github.com/hexops/mach/issues/60)?
```
slimsag, line 1428: accidental?

InKryption: Whoops; I'm not sure how that happened. I'm sure I pulled from main

slimsag:
> there are other places where glfw.dont_care is an accepted value to ask for "default behavior", such as in hints; should we also change those to be null-able, or leave them as is?

I am only concerned if we have missing feature parity with GLFW here. With window hints, we have default values for each struct field matching GLFW so that seems fine (I suspect it's unlikely `GLFW_DONT_CARE` would behave differently from the default value.)

With that in mind, do you have any concerns that we've missed feature parity somewhere?

InKryption:
No, if that's how we're handling it, then that sounds fine to me. I only ask because it adds some type-safety, such as I did for `Window.setMonitor`, where the 'refresh_rate' parameter was an `isize`, but was only supposed to take positive values, or `glfw.dont_care`. It's now a `?usize`, where `null` is equivalent to `glfw.dont_care`.

slimsag:
Ahhh I see, that makes sense. From my POV there's not substantial difference between having defaults and optionals. I do see why optionals give better type safety, but I also imagine they can make it a bit more annoying to work with the fields in practice (easier to know there's a value and if it wasn't specified it's the sane default, rather than need to handle a "not specified" case and figure that out for yourself)

I do think this means we should replace this line with the actual proper default value, instead of `glfw.dont_care`, however: https://github.com/hexops/mach/blob/b8c913cd6cf8e160fc1d680eac64f82f990a8e83/glfw/src/Window.zig#L192

Basically, `glfw.dont_care` should probably not be exposed publicly at all - we should use default values, and optionals, instead.

InKryption:
Sounds good and well to me; let me work on a commit for that. Then you can review it, and if all's well, should be ready for merging.

slimsag:
Most recent commit looks good.

I think you are right that we should allow those in the table to be nullable (but with the current default values still.) Although I am optimistic that the defaults will in practice be good enough for those.. I also don't know for sure that there is not some situation where null/dont_care is truly needed with these. The safe bet is to allow them to be nullable, agreed.

InKryption:
If everything looks good to you, merge away.

slimsag (approved):
Nice work!

PR #126, opened 2021-12-13
replaced isize, usize with i32, u32

### LINKED ISSUE #114: glfw: should use u32/i32/etc everywhere instead of usize/isize
Currently it produces an `isize` `width` and `height`, which matches the GLFW `int` type. But presumably a framebuffer is always positive and there's not a good reason to use signed integers?

AliChraghi:

closes #114

### DIFF
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -1645,7 +1645,7 @@ fn setFramebufferSizeCallbackWrapper(handle: ?*c.GLFWwindow, width: c_int, heigh
     internal_debug.assertInitialized();
     const window = from(handle.?) catch unreachable;
     const internal = window.getInternal();
-    internal.setFramebufferSizeCallback.?(window, @intCast(i32, width), @intCast(i32, height));
+    internal.setFramebufferSizeCallback.?(window, @intCast(u32, width), @intCast(u32, height));
 }
 
 /// Sets the framebuffer resize callback for the specified window.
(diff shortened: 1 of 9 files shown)


InKryption:
Everything looks pretty good; I think it's odd to cast to `u32` where indexes into arrays/slices/etc are expected, but I'm guessing that's just because of Search-And-Replace shenanigans?

AliChraghi:
idk, i just saw the issue title 
*should use u32/i32/etc **everywhere** instead of usize/isize* **xD**

InKryption:
Fair enough; not as though it matters too much anyway, so long as we're always indexing with an unsigned integer.

slimsag (changes requested):
Thanks a ton for sending this!

I marked a few places where we should continue to use `usize`, specifically in cases of slice indexing and where not using `usize` would result in undefined behavior.

Once those are updated, I'll merge this. Appreciate it!

glfw/src/Window.zig
```diff
@@ -48,16 +48,16 @@ pub const InternalUserPointer = struct {
...
-    setSizeCallback: ?fn (window: Window, width: isize, height: isize) void,
+    setPosCallback: ?fn (window: Window, xpos: i32, ypos: i32) void,
+    setSizeCallback: ?fn (window: Window, width: i32, height: i32) void,
     setCloseCallback: ?fn (window: Window) void,
     setRefreshCallback: ?fn (window: Window) void,
     setFocusCallback: ?fn (window: Window, focused: bool) void,
     setIconifyCallback: ?fn (window: Window, iconified: bool) void,
     setMaximizeCallback: ?fn (window: Window, maximized: bool) void,
-    setFramebufferSizeCallback: ?fn (window: Window, width: isize, height: isize) void,
+    setFramebufferSizeCallback: ?fn (window: Window, width: i32, height: i32) void,
```
slimsag, line 58: Ah, additionally, we should change `width` and `height` here to a `u32` not `i32`, since we determined in https://github.com/hexops/mach/issues/114#issuecomment-987910915 that it should be unsigned.

slimsag (approved):
Excellent, thanks again for this!

PR #129, opened 2021-12-15
gpu: Fix invalid parameter type of glfw framebuffer callback function

iddev5:

The dawn example currently wasnt compiling because the parameter type of glfw's framebuffer callback was incorrect. This was a leftover from #126 

Feel free to close it if already done locally.


slimsag (approved):
Thanks a bunch for sending this!

PR #134, opened 2021-12-22
glfw: add ziggified API of glfw3native.h

### LINKED ISSUE #131: glfw: should expose ziggified API for glfw3native.h
Currently everything in [glfw3native.h](https://github.com/hexops/mach/blob/main/glfw/upstream/glfw/include/GLFW/glfw3native.h) is unexposed by `mach-glfw`. We should find a nice way to expose a Zig API for these functions.

iddev5:

Closes #131

### DIFF
--- a/glfw/src/main.zig
+++ b/glfw/src/main.zig
@@ -25,6 +25,7 @@ pub const version = @import("version.zig");
 pub const VideoMode = @import("VideoMode.zig");
 pub const Window = @import("Window.zig");
 pub const Cursor = @import("Cursor.zig");
+pub const Native = @import("native.zig").Native;
 pub const Key = key.Key;
 
 pub usingnamespace @import("clipboard.zig");
--- a/glfw/src/native.zig
+++ b/glfw/src/native.zig
@@ -1,0 +1,412 @@
+//! Native access functions
+const std = @import("std");
+
+const Window = @import("Window.zig");
+const Monitor = @import("Monitor.zig");
+const Error = @import("errors.zig").Error;
+const getError = @import("errors.zig").getError;
+
+const internal_debug = @import("internal_debug.zig");
+
+const BackendOptions = struct {
+    win32: bool = false,
+    wgl: bool = false,
+    cocoa: bool = false,
+    nsgl: bool = false,
+    x11: bool = false,
+    glx: bool = false,
+    wayland: bool = false,
+    egl: bool = false,
+    osmesa: bool = false,
+};
+
+pub fn Native(comptime options: BackendOptions) type {
+    const native = @cImport({
+        @cDefine("GLFW_INCLUDE_VULKAN", "1");
+        @cInclude("GLFW/glfw3.h");
+
+        if (options.win32) @cDefine("GLFW_EXPOSE_NATIVE_WIN32", "1");
+        if (options.wgl) @cDefine("GLFW_EXPOSE_NATIVE_WGL", "1");
+        if (options.cocoa) @cDefine("GLFW_EXPOSE_NATIVE_COCOA", "1");
+        if (options.nsgl) @cDefine("GLFW_EXPOSE_NATIVE_NGSL", "1");
+        if (options.x11) @cDefine("GLFW_EXPOSE_NATIVE_X11", "1");
+        if (options.glx) @cDefine("GLFW_EXPOSE_NATIVE_GLX", "1");
+        if (options.wayland) @cDefine("GLFW_EXPOSE_NATIVE_WAYLAND", "1");
+        if (options.egl) @cDefine("GLFW_EXPOSE_NATIVE_EGL", "1");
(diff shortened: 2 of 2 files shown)


iddev5:
So something strange happened. For some reason addOptions is not being taken up properly. It is generating the proper options, adding the relevant option file as a package to the final library. But this package is not being found by the compiler inside the c.zig file.

So I have fall back to a simpler solution which is to always expose the native APIs. This works on linux except the OSMesa functions (related to Wayland). Further testing needed.

glfw/src/main.zig
```diff
@@ -33,6 +33,7 @@ pub usingnamespace @import("vulkan.zig");
 pub usingnamespace @import("time.zig");
 pub usingnamespace @import("hat.zig");
 pub usingnamespace @import("mod.zig");
+pub const native = @import("native.zig");
```
slimsag, line 36: Here's an idea, not 100% positive it'll work but I think so and may be a much nicer way to do this?

We could expose different imports here like `native_win32`, `native_wgl`, `native_cocoa`, `native_nsgl`, etc.

Inside of each of those, we could instead of `const c = @import("c.zig").c;` directly use `@cInclude("GLFW/glfw3native.h");` and `@cDefine` the right definitions.

Then only if someone references that code (`glfw.native_win32`, etc.) within theirs would Zig actually interpret the `@cInclude` in that `native_win32.zig` file, I believe, and so no special configuration or platform detection is needed (it's left up to the user)

glfw/src/native.zig
```diff
@@ -0,0 +1,290 @@
...
+}
+
+//
+
+pub fn getWGLContext(window: Window) std.os.windows.HGLRC {
+    internal_debug.assertInitialized();
+    const context = c.glfwGetWGLContext(window.handle);
+    getError() catch |err| return switch (err) {
+        Error.NotInitialized => unreachable,
+        Error.NoWindowContext => {}, // TODO
```
slimsag, line 48: We can/should fix these by returning an error set like this:

https://github.com/hexops/mach/blob/6866fce4ee3cb7c85f06adf603e416843725d60e/glfw/src/opengl.zig#L34

https://github.com/hexops/mach/blob/6866fce4ee3cb7c85f06adf603e416843725d60e/glfw/src/opengl.zig#L41

glfw/src/native.zig
```diff
@@ -0,0 +1,290 @@
...
+const c = @import("c.zig").c;
+
+const Window = @import("Window.zig");
+const Monitor = @import("Monitor.zig");
+const Error = @import("errors.zig").Error;
+const getError = @import("errors.zig").getError;
+
+const internal_debug = @import("internal_debug.zig");
+
+pub fn getWin32Adapter(monitor: Monitor) [*:0]const u8 {
```
slimsag, line 11: I'd also love to copy all the docs from glfw3native.h into this file as `///` doc comments on these functions, similar to what we do for others. Happy to do this boring work if you don't want to

glfw/src/native.zig
iddev5: I know, the PR is WIP :smile:

glfw/src/native.zig
iddev5: I will do that, dont worry, just want to figure out the actual work first

glfw/src/main.zig
iddev5: I actually tried something similar. The problem is that a @cImport containing @cInclude of glfw3native.h also needs to include glfw3.h, and doing that would create incompatibilities with c.zig's cImport.

One way to go around it is by @ptrCast-ing them, and that this guaranteed to work (despite looking ugly) but I would need some confirmation first.

To explain more, the way it is designed is, you use the API as follow:
```zig
const native = glfw.Native(.{ .x11 = true, .glx = true });
const xwindow = native.getX11Window(window); 
```

On the impl side:
```zig
pub fn Native(comptime options: Options) type {
    const native = @cImport({
        // include glfw3.h first because its needed
        if (options.x11) // define x11 flag
        // include glfw3native.h
    });
    return struct {
        pub fn getX11Window(window: Window) u32 {
            return @intCast(u32, native.glfwGetX11Window(@ptrCast(*native.GLFWwindow, window.handle)));
        }
    };
}
```

The reason why I am suggesting arbitary backend combination because certain backends like egl and osmesa works on all platform, so its up the the user to decide.

slimsag: Got it. I think I am OK with both your solution above, as well as @ptrCast'ing approach so whichever you are happier with sounds great to me!

iddev5:
Doc comments are not the best, but I think they can be improved later on.

iddev5:
I think I may have left out some *anyopaque casts, I will fix them tomorrow.

iddev5:
Done!

AliChraghi (approved):

slimsag (approved):
Looks great, thanks for this!

PR #135, opened 2021-12-22
add support for compiling for Wayland

### LINKED ISSUE #77: glfw: add support for cross-compiling for Wayland
Reported by @AliChraghi in https://github.com/hexops/mach/issues/75#issuecomment-969446408

we will need to add the right apt packages to the fetch.sh script here so our SDK has the headers/libs so we can cross compile for Wayland. https://github.com/hexops/sdk-linux-x86_64

It is worth noting that it may be worth waiting until GLFW's Wayland backend (which today is still very immature I think) improves, e.g. in the next version of GLFW there will be support for choosing between X11 and Wayland backends at runtime and so we should begin building for both by default. glfw/glfw#1958

AliChraghi:

current changes make zig able to compile for wayland. but there's a runtime error (`PlatformError`). i did't find the reason, maybe it only happen in my system. 
some commit's like `updating glfw to 3.3.6` are't belong to issue.
closes #77 and maybe needs to create another issue for runtime error


glfw/generate-wayland-protocols.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+# modified from https://github.com/go-gl/glfw/blob/master/scripts/generate-wayland-protocols.sh
```
slimsag, line 2: Instead of using a script like this that we'll have to maintain, etc. I think we should just find out what the right Ubuntu packages are and add them to this list:

https://github.com/hexops/sdk-linux-x86_64/blob/main/fetch.sh#L10-L27

Then we can rerun `fetch.sh` in that repository and commit the result, that way we have a super straightforward way to know we're getting a good version of these headers and we can easily manage other dependencies if GLFW should add them in the future.

slimsag:
Thanks for the PR! I'd like to separate out the commits in this PR into smaller ones:

* One for updating to GLFW version 3.3.6
* One for https://github.com/hexops/mach/pull/135#discussion_r774010720
* One for fixing Wayland compilation support

If that's okay with you.

I am also happy to do this for you if you don't want to (you'll still be listed as commit author)

PR #136, opened 2021-12-22
glfw: Update glfw to `3.3.6`

AliChraghi:


iddev5 (commented):
Why is it adding a workflow file containing cmake build jobs?

AliChraghi:
thanks for telling @iddev5, they need to be deleted

slimsag (approved):
LGTM, I left TODOs about updating Zig documentation strings but we can do that in follow-up PR.

PR #140, opened 2021-12-23
add wayland cross-compilation support

AliChraghi:
Fixes hexops/mach#77


glfw/src/sources_linux_wayland.c
```diff
@@ -2,3 +2,9 @@
 #include "wl_monitor.c"
 #include "wl_window.c"
 #include "wl_init.c"
+#include "wayland-viewporter-client-protocol.c"
```
slimsag, line 5:
```suggestion

// Note: These sources are generated by wayland-scanner, and distributed in sdk-linux-*
// See e.g. https://github.com/hexops/sdk-linux-x86_64/blob/main/generate-wayland.sh 
#include "wayland-viewporter-client-protocol.c"
```

slimsag (approved):

PR #142, opened 2021-12-24
gpu-dawn: Fix building after the rename of gpu -> gpu-dawn

iddev5:

I noticed building of dawn was broken as a result of #139 . This small PR fixes it

Builds locally for me


iddev5:
Hmm I feel like gpu-dawn needs a CI now

slimsag (approved):
Thanks!

PR #147, opened 2021-12-24
glfw: expose glfwSetErrorCallback for retrieving optional error descriptions

### LINKED ISSUE #144: glfw: need a way to get error descriptions
@AliChraghi brought up a really good point in the Matrix chat: `mach/glfw` currently eats error descriptions, and sometimes GLFW's error types are not sufficient to debug the cause:

image

https://github.com/hexops/mach/blob/38cab2e423b14bb1c8a6273a9387be6b0f9d6c38/glfw/src/errors.zig#L125

currently we are just calling glfwGetError and we pass in a null pointer for the description, so GLFW doesn't give it to us.

Open questions:

* What level of detail is in GLFW error descriptions? Specifically of concern would be if Zig GLFW apps in the wild could produce these errors and we would need such error descriptions from users in order to debug the cause..
* Is there a sane way we could return error messages with errors? Should we?
* Would a debug option (env var?) be better? e.g. when set, print error descriptions to stderr.

AliChraghi:

closes #144

### DIFF
--- a/glfw/src/errors.zig
+++ b/glfw/src/errors.zig
@@ -1,5 +1,6 @@
 //! Errors
 
+const testing = @import("std").testing;
 const c = @import("c.zig").c;
 
 /// Errors that GLFW can produce.
@@ -87,6 +88,8 @@ pub const Error = error{
     NoWindowContext,
 };
 
+pub const ErrorCallbackFn = fn (c_int, [*c]const u8) callconv(.C) void;
+
 fn convertError(e: c_int) Error!void {
     return switch (e) {
         c.GLFW_NO_ERROR => {},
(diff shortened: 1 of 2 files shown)


AliChraghi:
we need to know log level or just use `log.warn`

iddev5:
Sorry for interrupting and I know this is meant to be a final PR but what if instead of printing the error like we, we instead expose the ``glfwSetErrorCallback()`` function in a ziggified API for the user to print it however they like?

AliChraghi:
i agree.

iddev5 (approved):

slimsag (approved):

PR #152, opened 2022-01-02
glfw: Tidy up UserPointer access

InKryption:
This change both restricts and clarifies the mutability/nullability of the pointers, and replaces the explicitly-typed pointer usage in setUserPointer with ?*anyopaque, since it now, as of being renamed from c_void, more simply communicates the intent of taking any pointer type.


AliChraghi (approved):

InKryption:
Pardon the force pushes, been a while, and I keep messing up the commit message

slimsag (approved):
Nice! Thanks for this!

PR #155, opened 2022-01-10
glfw: Use `anyopaque` instead of `opaque{}`

InKryption:
Using an inline `opaque{}` type forces the use of `@typeInfo` to cast to the specific type of the parameter.

### DIFF
--- a/glfw/src/vulkan.zig
+++ b/glfw/src/vulkan.zig
@@ -143,8 +143,8 @@ pub fn getInstanceProcAddress(vk_instance: ?*anyopaque, proc_name: [*:0]const u8
 ///
 /// see also: vulkan_present
 pub inline fn getPhysicalDevicePresentationSupport(
-    vk_instance: *opaque {},
-    vk_physical_device: *opaque {},
+    vk_instance: *anyopaque,
+    vk_physical_device: *anyopaque,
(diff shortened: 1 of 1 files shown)


AliChraghi (approved):

glfw/src/vulkan.zig
```diff
@@ -109,7 +109,7 @@ pub const VKProc = fn () callconv(.C) void;
 /// @pointer_lifetime The returned function pointer is valid until the library is terminated.
 ///
 /// @thread_safety This function may be called from any thread.
-pub fn getInstanceProcAddress(vk_instance: ?*opaque {}, proc_name: [*:0]const u8) callconv(.C) ?VKProc {
+pub fn getInstanceProcAddress(vk_instance: ?*anyopaque, proc_name: [*:0]const u8) callconv(.C) ?VKProc {
```
slimsag, line 112: Do you mind updating the other two instances of `*opaque {}` below in the `getPhysicalDevicePresentationSupport` method?

InKryption: Good spot; just pushed this because the only thing I was using was this first one, slipped my mind to consider the other two instances.

slimsag (approved):
Perfect, thanks a ton!

PR #158, opened 2022-02-08
glfw: improve `setErrorCallback`

leecannon:
This PR changes `setErrorCallback` to take a callback with nice zig types.

As this callback can be set before `glfwInit` is called the method of storing function pointers on `InternalUserPointer` cannot be used.

To make it work the function provided by the user must be comptime known allowing us to construct a c wrapper function inline. 


slimsag (approved):
Wow this is a great approach, thanks for this!

PR #159, opened 2022-02-08
glfw: dont call `getError` unless we need to

leecannon:
A large chunk of glfw functions inform us of success using their return values, that means that calling `getError` is unnecessary in those cases.

The cost of `getError` (for the non-error case) is more than the single branch checking the return value:

1. call to `glfwGetError`
    1. check `description` pointer
    2. platform tls lookup
    3. check error is not `GLFW_NO_ERROR`
2. switch from int to error in `convertError`


slimsag (approved):
Thanks so much for sending this! Great idea

PR #160, opened 2022-02-08
glfw: dont use `@errSetCast`

leecannon:
In safe build modes `@errSetCast` emits *runtime* checks to validate the error is in the destination errorset.

As switches do errorset narrowing automatically `@errSetCast` is unnecessary and validation of the error being in the errorset occurs at compile time.


slimsag (approved):
Love this, thanks so much! Now I want a Zig linter for this type of thing :)

PR #162, opened 2022-02-09
glfw: use comptime magic to remove `InternalUserPointer` and associated overhead

leecannon:
This PR changes the way glfw callback functions are implemented.

## Improvements

**Before**
Inside Window.zig `from` there is an allocation of a `InternalUserPointer` with an allocator not explicitly provided by the caller.

**After**
No hidden allocation anywhere. (atleast in the zig wrapper's code)

**Before**
Due to using the user pointer functionality provided by glfw to carry around function callbacks, the API to work with the end users actual user pointer generates more code.

**After**
`getUserPointer` & `setUserPointer` are trivial passthrough functions.

**Before**
When an event occurs that the user has set a callback for the below takes place before the users callback code is reached:
- call through a runtime function pointer from glfw to a callback wrapper
- call back into glfw to get the userPointer and check if its null (Window.zig `from`)
- null check on the user provided function pointer
- cast/transform the c callbacks parameters into nice zig ones
- call through a runtime function pointer from the callback wrapper to the users callback function

**After**
When an event occurs that the user has set a callback for the below takes place before the users callback code is reached:
- call through a runtime function pointer from glfw to the users callback function
- cast/transform the c callbacks parameters into nice zig ones

There is no second call through a runtime function pointer due to a very nice thing about functions that are taken as comptime arguments, they are statically determined at compile time and calling them results in the same codegen as calling a function directly. Due to this we can even force the function to be fully inlined essentially turning the c wrapper function into a prelude of the users callback function with no additional function call overhead.

## API Change
The change to take the functions as comptime arguments does prevent runtime known funcion pointers from being provided, however in almost every situation a truly runtime function pointer is rare.
The most common reason a function pointer is runtime known is due to deciding between different implementations based on a runtime value, however this can easily work with the new API:
```zig
switch (runtime_value) {
    .one => window.setPosCallback(posCallbackOne),
    .two => window.setPosCallback(posCallbackTwo),
    .three => window.setPosCallback(posCallbackThree),
}
```

Or the user could decide to install a callback that calls a runtime known function pointer, this would be very similar to how the current API is implemented, however the choice to do this would be in the users hands.

### DIFF
--- a/glfw/src/Joystick.zig
+++ b/glfw/src/Joystick.zig
@@ -352,12 +352,6 @@ pub const Event = enum(c_int) {
     disconnected = c.GLFW_DISCONNECTED,
 };
 
-var _callback: ?fn (joystick: Joystick, event: Event) void = null;
-
-fn callbackWrapper(jid: c_int, event: c_int) callconv(.C) void {
-    _callback.?(Joystick{ .jid = @intToEnum(Joystick.Id, jid) }, @intToEnum(Event, event));
-}
-
 /// Sets the joystick configuration callback.
 ///
 /// This function sets the joystick configuration callback, or removes the currently set callback.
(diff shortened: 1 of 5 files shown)


InKryption:
I like this change. I squinted at this part of the library for a bit, but never thought of a solution like this, aside from maybe allowing the user to provide an allocator.

PR #165, opened 2022-02-14
Update the year in the license file

C-BJ:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2021, Hexops Contributors (given via the Git commit history).
+Copyright 2021-2022, Hexops Contributors (given via the Git commit history).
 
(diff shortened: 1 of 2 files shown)


slimsag:
Hi, sorry but I don't really see a need to do this.

Copyright law in US and EU is the lifetime of the author plus like 70 years. Maybe it makes sense to do in China where copyright expires after 50 years apparently.. but in any case, the code is given away under two very permissive open source licenses so I'm not really worried about this every year.

It's also annoying/tedious to do this everywhere, in every repository, etc. every year 😜 For example, there are multiple copies of these licenses in this repo you missed by accident in this update.

I do appreciate the PR, thank you for taking the time to look / contribute! Hope it's okay I turn this change down. :)

PR #181, opened 2022-03-14
glfw: Enable Gyro support (to more easily add this as a dependency)

BratishkaErik:

### DIFF
--- a/glfw/.gitignore
+++ b/glfw/.gitignore
@@ -16,3 +16,6 @@ zig-out/
 /build/
 /build-*/
 /docgen_tmp/
+.gyro/
+deps.zig
+gyro.lock
--- a/glfw/gyro.zzz
+++ b/glfw/gyro.zzz
@@ -1,0 +1,26 @@
+pkgs:
+  mach-glfw:
+    description: Ziggified GLFW bindings
+    source_url: https://github.com/hexops/mach-glfw
+    version: 0.0.0
+    root: src/main.zig
+    tags:
+      bindings
+    files:
+      README.md
+      LICENSE
+      LICENSE-APACHE
+      LICENSE-MIT
+
+  mach-glfw-build:
+    description: build.zig for GLFW linking
+    source_url: https://github.com/hexops/mach-glfw
+    version: 0.0.0
+    root: build.zig
+    tags:
+      build
+    files:
+      README.md
+      LICENSE
+      LICENSE-APACHE
+      LICENSE-MIT


slimsag:
I'm not super sure what to do here.

[Gyro](https://github.com/mattnite/gyro) and [zigmod](https://github.com/nektro/zigmod) are the two popular unofficial package managers for Zig, and an official one is on the way. Many use submodules today as a workaround. I personally will only be using the official package manager, and in the meantime submodules.

The problem comes in that this introduces a bunch of questions I don't really know how to answer:

* If we support Gyro, should we not also support zigmod?
* Once the official package manager comes out, can we remove Gyro and zigmod? That doesn't seem nice / like a good citizen, and I imagine would leave us with an "outdated" package on these indexes despite the library otherwise being maintained.
    * Do we then commit to maintaining the gyro/zigmod package definition forever going forward?
* How do we update & keep the Gyro file / zigmod file up to date? Since I won't be familiar with these tools, it's hard for me to do this. And if a third package manager comes out.. where do I stop?

I totally understand if all these sound like exaggerated concerns, you're just a user of Gyro and want this merged for easier usage on your end.. I'm just worried about the long term impact this will have, and would like to wait a bit longer until package management in Zig calms down a bit before deciding what to do exactly.

BratishkaErik:
> I'm not super sure what to do here.
> 
> [Gyro](https://github.com/mattnite/gyro) and [zigmod](https://github.com/nektro/zigmod) are the two popular unofficial package managers for Zig, and an official one is on the way. Many use submodules today as a workaround. I personally will only be using the official package manager, and in the meantime submodules.
> 
> The problem comes in that this introduces a bunch of questions I don't really know how to answer:
> 
>     * If we support Gyro, should we not also support zigmod?
> 
>     * Once the official package manager comes out, can we remove Gyro and zigmod? That doesn't seem nice / like a good citizen, and I imagine would leave us with an "outdated" package on these indexes despite the library otherwise being maintained.
>       
>       * Do we then commit to maintaining the gyro/zigmod package definition forever going forward?
> 
>     * How do we update & keep the Gyro file / zigmod file up to date? Since I won't be familiar with these tools, it's hard for me to do this. And if a third package manager comes out.. where do I stop?
> 
> 
> I totally understand if all these sound like exaggerated concerns, you're just a user of Gyro and want this merged for easier usage on your end.. I'm just worried about the long term impact this will have, and would like to wait a bit longer until package management in Zig calms down a bit before deciding what to do exactly.

Ok, new idea :)

PR #184, opened 2022-03-19
glfw: add instructions for Gyro in README.md

BratishkaErik:
Instead of #181 


slimsag (approved):
Beautiful compromise, thank you a ton for this!

PR #200, opened 2022-03-31
gpu-dawn: build.zig needs to reference thisDir for C file

mitchellh:
This allows gpu-dawn to be vendored into a project when building from source.


slimsag (approved):
Thank you!

PR #206, opened 2022-04-06
Tweak build.zigs to make using app library slightly easier

silversquirl:
There are a few changes in one PR here:
- Add explicit `std.build.Pkg` types to `pkg` variables in `build.zig`s
- Publicize `gpu` and `glfw` from the toplevel `build.zig` to make it easier to use those from projects importing that
- Run `zig fmt` on the toplevel build.zig (one tiny whitespace change)

---


slimsag (approved):
Thanks a ton for this, @silversquirl !

PR #209, opened 2022-04-08
Fixes and API tweaks for mach/gpu

silversquirl:
This PR fixes a number of issues I ran into while trying to write some stuff using mach/gpu.
A few of them are simple bugs, but a few are more stylistic choices - feel free to reject or change those! :)


slimsag (approved):
Beautiful, thanks so much for this! I like all of these changes :)

PR #213, opened 2022-04-10
Added helper functions to BindGroupLayout.Entry

michal-z:
Also added default values to `*.BindingLayout` structures like this is done in `webgpu_cpp.h`.
Without those methods and default values the code to init `BindGroupLayout` is very verbose.

With this PR initialization is much cleaner:

```zig
    const desc = BindGroupLayout.Descriptor{
        .entries = &.{
            BindGroupLayout.Entry.buffer(0, .{ .vertex = true }, .uniform, true, 0),
            BindGroupLayout.Entry.sampler(1, .{ .vertex = true }, .filtering),
            BindGroupLayout.Entry.texture(2, .{ .fragment = true }, .float, .dimension_2d, false),
            BindGroupLayout.Entry.storageTexture(3, .{ .fragment = true }, .none, .rgba32_float, .dimension_2d),
        },
    };
```

### DIFF
--- a/gpu/src/BindGroupLayout.zig
+++ b/gpu/src/BindGroupLayout.zig
@@ -1,7 +1,9 @@
 const Buffer = @import("Buffer.zig");
 const Sampler = @import("Sampler.zig");
 const Texture = @import("Texture.zig");
+const TextureView = @import("TextureView.zig");
 const StorageTextureBindingLayout = @import("structs.zig").StorageTextureBindingLayout;
+const StorageTextureAccess = @import("enums.zig").StorageTextureAccess;
 const ShaderStage = @import("enums.zig").ShaderStage;
 
 const BindGroupLayout = @This();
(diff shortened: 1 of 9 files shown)


gpu/src/Buffer.zig
```diff
@@ -103,9 +103,9 @@ pub const BindingType = enum(u32) {
 
 pub const BindingLayout = extern struct {
     reserved: ?*anyopaque = null,
-    type: BindingType,
-    has_dynamic_offset: bool,
-    min_binding_size: u64,
+    type: BindingType = .none,
```
slimsag, line 106: I think we should have this default to `.uniform` since that's what it's defined as in the spec:

image

https://www.w3.org/TR/webgpu/#dictdef-gpubufferbindinglayout

gpu/src/Sampler.zig
```diff
@@ -36,7 +36,7 @@ pub const BindingType = enum(u32) {
 
 pub const BindingLayout = extern struct {
     reserved: ?*anyopaque = null,
-    type: BindingType,
+    type: BindingType = .none,
```
slimsag, line 39: Similarly, should default to `.filtering`

image

https://www.w3.org/TR/webgpu/#dictdef-gpusamplerbindinglayout

gpu/src/Texture.zig
```diff
@@ -211,9 +211,9 @@ pub const SampleType = enum(u32) {
 
 pub const BindingLayout = extern struct {
     reserved: ?*anyopaque = null,
-    sample_type: SampleType,
-    view_dimension: TextureView.Dimension,
-    multisampled: bool,
+    sample_type: SampleType = .none,
+    view_dimension: TextureView.Dimension = .dimension_none,
+    multisampled: bool = false,
```
slimsag, line 216: `.float` and `.dimension_2d`

image

gpu/src/structs.zig
```diff
@@ -39,9 +39,9 @@ pub const PrimitiveState = struct {
 
 pub const StorageTextureBindingLayout = extern struct {
     reserved: ?*anyopaque = null,
-    access: StorageTextureAccess,
-    format: Texture.Format,
-    view_dimension: TextureView.Dimension,
+    access: StorageTextureAccess = .none,
+    format: Texture.Format = .none,
+    view_dimension: TextureView.Dimension = .dimension_none,
```
slimsag, line 44: `.dimension_2d`

gpu/src/structs.zig
```diff
@@ -39,9 +39,9 @@ pub const PrimitiveState = struct {
 
 pub const StorageTextureBindingLayout = extern struct {
     reserved: ?*anyopaque = null,
-    access: StorageTextureAccess,
-    format: Texture.Format,
-    view_dimension: TextureView.Dimension,
+    access: StorageTextureAccess = .none,
```
slimsag, line 42: `.write_only`

gpu/src/structs.zig
```diff
@@ -39,9 +39,9 @@ pub const PrimitiveState = struct {
...
+    format: Texture.Format = .none,
```
slimsag, line 43: The spec has this to say - I think we should make this required for now, no default `.none`:

image

gpu/src/BindGroupLayout.zig
```diff
@@ -38,10 +40,72 @@ pub const Entry = extern struct {
...
     binding: u32,
     visibility: ShaderStage,
-    buffer: Buffer.BindingLayout,
-    sampler: Sampler.BindingLayout,
-    texture: Texture.BindingLayout,
-    storage_texture: StorageTextureBindingLayout,
+    buffer: Buffer.BindingLayout = .{},
+    sampler: Sampler.BindingLayout = .{},
+    texture: Texture.BindingLayout = .{},
+    storage_texture: StorageTextureBindingLayout = .{},
```
slimsag, line 46: I actually think these should be optional values, according to the spec, not default to zero values.

image

gpu/src/BindGroupLayout.zig
```diff
@@ -38,10 +40,72 @@ pub const Entry = extern struct {
...
+
+    pub fn buffer(
```
slimsag, line 48: Could add comments to these to make it clear they're helpers:

```
/// Helper to create a buffer BindGroupLayout.
```

slimsag:
LGTM otherwise!

gpu/src/BindGroupLayout.zig
michal-z: `BindingLayoutGroup.Entry' is an `extern struct` - we can't have optional field for ABI compatibility.

michal-z:
Note that `BindGroupLayout.Entry` represents an union of various binding types. The idea is that all bindings are undefined by default and then the user sets only the one that is needed.

https://www.w3.org/TR/webgpu/#dictdef-gpubindgrouplayoutentry

gpu/src/BindGroupLayout.zig
slimsag: Ahh, you're right!

slimsag (approved):
Perfect, thanks again!

PR #222, opened 2022-04-14
Add cube and some more examples, zmath for testing

PiergiorgioZagaria:


ifreund (changes requested):
Something you've done has removed the executable permission from the scripts in `dev/`. They should remain 755.

michal-z:
One note/suggestions, zmath uses row-major matrices, so the order of multiplication is:

`v * model_mat * view_mat * proj_mat`

as you are doing on the CPU side. To be consistent, it is recommended to do the transposition of the final `mvp` matrix just before passing it to the shader - it lets you use the same order of multiplication in the shader, so:

`output.position_clip = vec4(position, 1.0) * ubo.transform;`

PiergiorgioZagaria:
> Something you've done has removed the executable permission from the scripts in `dev/` and elsewhere. They should remain 755.

How do I change the permissions?

PiergiorgioZagaria:
> One note/suggestions, zmath uses row-major matrices, so the order of multiplication is:
> 
> `v * model_mat * view_mat * proj_mat`
> 
> as you are doing on the CPU side. To be consistent, it is recommended to do the transposition of the final `mvp` matrix just before passing it to the shader - it lets you use the same order of multiplication in the shader, so:
> 
> `output.position_clip = vec4(position, 1.0) * ubo.transform;`

I have not really thought much about the matrix multiplication, but based on what you wrote, I think I should just swap the vec4 and ubo.transform in the vertex shader right? (Please tell me if I misunderstood.)
But doing so doesn't give the correct view (tested).

michal-z:
Yes, but you also need to `zm.transpose()` the final `mvp` matrix, so in your case, I would do:

```
        const mvp = zm.mul(zm.mul(model, view), proj);
        const ubo = UniformBufferObject{
            .mat = zm.transpose(mvp),
            .time = time,
        };
```

PiergiorgioZagaria:
I think the matrix multiplications and the permissions are ok now

ifreund:
Indeed, the permissions look good, I did just notice that you seem to have deleted the `DirectXShaderCompiler` and `dawn` submodules from `gpu-dawn/libs/` however.

PiergiorgioZagaria:
I think that didn't fix the submodules,

build.zig
```diff
@@ -23,14 +23,31 @@ pub fn build(b: *std.build.Builder) void {
...
+
+        const example_run_cmd = example.run();
+        example_run_cmd.step.dependOn(b.getInstallStep());
+        const example_run_step = b.step("run-example-" ++ name, "Run the example");
+        example_run_step.dependOn(&example_run_cmd.step);
+    }
 
-    inline for ([_][] const u8{"triangle", "boids"}) |name| {
+    const zmath_pkg = @import("zmath/build.zig").pkg;
+    inline for ([_][]const u8{ "mandelbrot", "fragment-shader", "rotating-cube", "two-cubes" }) |name| {
```
iddev5, line 43: Would be better if both the example collection can be combined, maybe with an additional needs_zmath field or with a []Pkg additional_deps field. This would require switching to use a struct to represent the examples

examples/mandelbrot/main.zig
```diff
@@ -0,0 +1,252 @@
...
+    {
+        const time = @intToFloat(f32, timer.read()) / @as(f32, std.time.ns_per_s);
+        const model = zm.rotationZ(0);
+        const view = zm.lookAtRh(
+            zm.f32x4(0, 2, 0, 1),
+            zm.f32x4(0, 0, 0, 1),
+            zm.f32x4(0, 0, 1, 0),
+        );
+        const proj = zm.perspectiveFovRh(
+            (std.math.pi / 7.0), //(std.math.sin(time) * 7 + 9)),
```
iddev5, line 223: Stray commented out code?

examples/rotating-cube/main.zig
```diff
@@ -0,0 +1,283 @@
...
+
+    .{ .pos = .{ 1, -1, -1, 1 }, .col = .{ 1, 0, 0, 1 }, .uv = .{ 1, 1 } },
+    .{ .pos = .{ -1, -1, -1, 1 }, .col = .{ 0, 0, 0, 1 }, .uv = .{ 0, 1 } },
+    .{ .pos = .{ -1, 1, -1, 1 }, .col = .{ 0, 1, 0, 1 }, .uv = .{ 0, 0 } },
+    .{ .pos = .{ 1, 1, -1, 1 }, .col = .{ 1, 1, 0, 1 }, .uv = .{ 1, 0 } },
+    .{ .pos = .{ 1, -1, -1, 1 }, .col = .{ 1, 0, 0, 1 }, .uv = .{ 1, 1 } },
+    .{ .pos = .{ -1, 1, -1, 1 }, .col = .{ 0, 1, 0, 1 }, .uv = .{ 0, 0 } },
+};
+
+// If you want to use indices for the cube
```
iddev5, line 59: Instead of having all these commented out, I wonder if this would be better to rather have them inside a build option (using b.addOptions and family) cc: slimsag

build.zig
PiergiorgioZagaria: Seems like the best approach, you are right

examples/mandelbrot/main.zig
PiergiorgioZagaria: I felt like leaving It so that when viewing the example you had an Easy way to zoom in, but i think I should probably add later the zoom from mouse input instead of time

examples/rotating-cube/main.zig
PiergiorgioZagaria: I thought these were more like templates to read and try stuff on. So that you could add, remove and uncomment stuff to test the effects It had on the program on the fly.

Doesn't really feel like a build option.

Maybe I could separate the vertices in a cube_mesh.zig file where there are both vertices, and vertices to index, so that you can import the One you want easily and don't have to Copy that each time you want to draw a cube

iddev5: I think you're right, its good as is. Too much engineering and examples would lose their purpose

slimsag: Agreed, we should strive for examples to be a single Zig file in general (+some assets, WGSL shaders, etc.)

slimsag:
@PiergiorgioZagaria really great work here, thanks so much for sending this.

hsome thoughts:

* let's put `zmath` into a separate repository for now (can be a temporary one like `github.com/PiergiorgioZagaria/zmath` or something), I don't want to commit the entire source file here, and I also don't want to pull in the entire zig-gamedev repository due to size. Then we could add it as a submodule inside the `examples/` directory, at `examples/libs/zmath`. (just temporary until Zig has a proper package manager we can rely on.)

* Would you mind sending separate PRs for each example? I think it will be easier to review/merge this way. Smaller is better :) If this sounds like a lot of work, no worries, I am happy to do it for you if you prefer

* It seems like there are still some changes to unrelated files, `gpu-dawn/libs` and `.gitmodules` - we should undo those.

examples/mandelbrot/main.zig
slimsag: I like it 👍 It'd be great to add zoom/panning via mouse input in the future (not as part of this PR.)

slimsag:
Also, is it cool if I list you under the contributors section [on the website](https://hexops.com/mach/)?

PiergiorgioZagaria:
>     * Would you mind sending separate PRs for each example? I think it will be easier to review/merge this way. Smaller is better :) If this sounds like a lot of work, no worries, I am happy to do it for you if you prefer

Ok so I close this PR now and add each example separately.


> Also, is it cool if I list you under the contributors section [on the website](https://hexops.com/mach/)?

Fine by me, though it doesn't really feel like I have done much

slimsag:
Hey, four great examples ported over is still four great examples :)

PiergiorgioZagaria:
> Hey, four great examples ported over is still four great examples :)

The examples that were ported are just 2, (and they still lack the depth stencil part even if they work), the other two are just examples of what can be done with fragment shaders that I wanted to add to give something that might also catch the eye. I was also thinking of cutting them and just make a template for working and testing fragment shaders, since I don't know of tools like shadertoy or shadered that support wgsl.
 
Anyway I created the https://github.com/PiergiorgioZagaria/zmath repo. I will wait for the submodule to be added and then make the PRs

slimsag:
> The examples that were ported are just 2, (and they still lack the depth stencil part even if they work)

Oh nice, I thought they were all ported! Even more reason to credit you :) Would you mind listing in https://github.com/hexops/mach/blob/main/examples/LICENSE which are ported over?

> I will wait for the submodule to be added and then make the PRs

Cool, I'll do this right now.

PiergiorgioZagaria:
Shouldn't we also list the triangle example? I noticed now that it is not mentioned in  https://github.com/hexops/mach/blob/main/examples/LICENSE

slimsag:
Only if they're ported from the webgpu-samples repository. The triangle example is not one that was ported over from there.

PR #223, opened 2022-04-16
fix getMappedRange size alignment

d3m1gd:
+ same for getConstMappedRange
+ mirror align fix for getConstMappedRange

rationale:

user needs to align buffer on creation, as in

    gpu.Buffer.Descriptor{ .size = size, ... }

so `size == sizeOf(T) * len` might not always hold true.

on `getMappedRange` request, there is no way to specify
that size was manually aligned, and using the function directly
with realigned size gives extremely unhelpful error:

    thread 254201 panic: attempt to use null value
    .../mach/gpu/src/NativeInstance.zig:1721:42: 0x480747
    in .gpu.NativeInstance.buffer_vtable.getMappedRange (game)
            return @ptrCast([*c]u8, range.?)[0..size];
                                          ^

it seems, requested size must match size specified on creation exactly.

so this commit does exactly that. aligns buffer size to 4 - something
that user must have done already if this point is reached.

aligned buffer sizes stay unchanged.

the whole point is to reduce mental load for the user and prevent hacks
and workarounds for corner cases. original issue originated with
having u16 index buffer array with odd number of elements, which is a
perfectly valid thing to have.

it would be best to also auto align size on buffer creation to not force
user to know such obscure implementation details and learn magic numbers.

### DIFF
--- a/gpu/src/Buffer.zig
+++ b/gpu/src/Buffer.zig
@@ -37,12 +37,14 @@ pub inline fn destroy(buf: Buffer) void {
 }
 
 pub inline fn getConstMappedRange(buf: Buffer, comptime T: type, offset: usize, len: usize) []const T {
-    const data = buf.vtable.getConstMappedRange(buf.ptr, offset, @sizeOf(T) * len);
-    return @ptrCast([*]const T, data.ptr)[0..len];
+    const size = @sizeOf(T) * len;
+    const data = buf.vtable.getConstMappedRange(buf.ptr, offset, size + size % 4);
+    return @ptrCast([*]const T, @alignCast(@alignOf(T), data.ptr))[0..len];
 }
 
 pub inline fn getMappedRange(buf: Buffer, comptime T: type, offset: usize, len: usize) []T {
-    const data = buf.vtable.getMappedRange(buf.ptr, offset, @sizeOf(T) * len);
+    const size = @sizeOf(T) * len;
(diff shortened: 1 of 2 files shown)


d3m1gd:
technically speaking, buffer size must be at least 4, so might want to `size.max(4)` or something similar, but i'm not sure how practical this is - creating buffer with size less than 4.

d3m1gd:
also, 4 is a magic number, so might define it more officially. wgpu.rs defines it as a constant `COPY_BUFFER_ALIGNMENT` for the whole crate

slimsag:
Thank you @d3m1gd! This looks like a good improvement over the status quo today, and so I've immediately merged this.

I think that we may need to do more here to be completely correct, though:

* `COPY_BUFFER_ALIGNMENT` is not exposed through https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h and so this makes me wonder if `4` is always the right alignment or not.
* There exist [`minUniformBufferOffsetAlignment` and `minStorageBufferOffsetAlignment`](https://github.com/webgpu-native/webgpu-headers/blob/b46ca3c39249c537b2b421a37cf06b0d20bbf2c1/webgpu.h#L753-L754), I'm not sure how they relate yet.
* We should indeed handle all buffer sizes in some reasonable way.

I will dig into these further / reach out to the Dawn authors for clarification, unless you beat me to it :)

PR #226, opened 2022-04-17
mach: set error message to be shown on glfw error

iddev5:
This uses glfw.setErrorCallback to set a global callback in mach's App which would work across all mach applications

How would this behave?
- The callback would show the error message first and then the regular behavior will follow: app would shutdown with an error trace

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -52,6 +52,11 @@ pub const Options = struct {
     power_preference: gpu.PowerPreference = .none,
 };
 
+/// Default GLFW error handling callback
+fn glfwErrorCallback(error_code: glfw.Error, description: [:0]const u8) void {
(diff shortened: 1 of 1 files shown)


src/main.zig
```diff
@@ -79,6 +84,7 @@ pub fn App(comptime Context: type, comptime config: AppConfig) type {
             const backend_type = try util.detectBackendType(allocator);
 
             try glfw.init(.{});
+            glfw.setErrorCallback(glfwErrorCallback);
```
slimsag, line 87: Should we set this before init? Since init can fail

iddev5: Done

slimsag (approved):

PR #233, opened 2022-04-18
Fixed ErrorCallback.init() alignment

PiergiorgioZagaria:

### DIFF
--- a/gpu/src/Adapter.zig
+++ b/gpu/src/Adapter.zig
@@ -164,7 +164,7 @@ pub const RequestDeviceCallback = struct {
     ) RequestDeviceCallback {
         const erased = (struct {
             pub inline fn erased(type_erased_ctx: *anyopaque, response: RequestDeviceResponse) void {
-                callback(if (Context == void) {} else @ptrCast(Context, @alignCast(@alignOf(Context), type_erased_ctx)), response);
+                callback(if (Context == void) {} else @ptrCast(Context, @alignCast(std.meta.alignment(Context), type_erased_ctx)), response);
             }
         }).erased;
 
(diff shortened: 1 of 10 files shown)


slimsag:
@PiergiorgioZagaria do you mind updating all locations with a bulk find-and-replace in your editor? Fine to do it in this PR. Shouldn't be too hard, no need to test each one or anything. Seems like there are a lot of locations we need to update:

https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/hexops/mach%24+callback%28if+%28Context+%3D%3D+void%29&patternType=literal

PiergiorgioZagaria:
Oh I didn't think about that, will do it right now

PiergiorgioZagaria:
I checked and there are more parts where it appears,
https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/hexops/mach%24+%40ptrCast%28Context%2C+%40alignCast%28%40alignOf%28Context%29&patternType=literal

PiergiorgioZagaria:
Plus some more that don't use context but also ptrCast(pointer, alignCast(alignOf(Pointer

slimsag:
Good catch, I think we only need to do it in `Callback` struct types' init methods.

A find and replace for `@ptrCast(Context, @alignCast(@alignOf(Context)` should be good enough I think.

(unrelated, looks like some of these locations need to be updated with `if (Context == void) {} else @ptrCast` too for consistency / `void` context support.)

slimsag (approved):
Looks great, thanks so much again!

PiergiorgioZagaria:
No prob

PR #245, opened 2022-04-20
shaderexp: add initial shader explorer tool

PiergiorgioZagaria:

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -50,6 +50,21 @@ pub fn build(b: *std.build.Builder) void {
         const example_run_step = b.step("run-example-" ++ example.name, "Run the example");
         example_run_step.dependOn(&example_run_cmd.step);
     }
+
+    // Is it ok to leave this here?
+    const shaderexp_exe = b.addExecutable("shaderexp", "shaderexp/main.zig");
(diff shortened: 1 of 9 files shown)


slimsag:
Very nice work! I pushed a few small changes, one to improve error handling, and another to fix an issue I ran into (if shader doesn't have `main` entrypoint it would crash):

```
mach: found Metal backend on Discrete GPU adapter: Apple M1, Metal driver on macOS Version 12.1 (Build 21C52)
info: The fragment shader has been changed
gpu: validation error: Entry point "main" doesn't exist in the shader module [ShaderModule "my fragment shader"].
 - While validating fragment stage (module: [ShaderModule "my fragment shader"], entryPoint: main).
 - While validating fragment state.
 - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor]).
```

Quite nice work, looks very cool! Did you make all the shaders yourself, or reuse some from somewhere else? (want to make sure we credit properly if so)

slimsag (approved):

PiergiorgioZagaria:
The raymarching one is a slight modification of the one in https://www.shadertoy.com/view/XlGBW3.
The rest are mine.

PR #247, opened 2022-04-21
examples: add advanced comptime generated texture + Blinn-Phong lighting example

d3m1gd:


slimsag:
This is great! I pushed a few changes:

* Renamed to `advanced-gen-texture-light`, since it's near ~1,000 lines (probably not best for beginner to start with immediately) and to make it clear it's generated texture / not loaded from disk.
* Fixed issues with crashing on macOS (use framebuffer size not window size), and crash during resizing (must recreate depth buffer if window resized)
* Added more instances by default, made camera rotate slowly by default
* Made movement independent of frame-rate using `app.delta_time`. I noticed it would go really fast with `App(...).init(.{.vsync = .none})` vsync turned off and during startup, so I fixed this.

Cool if I list you as a contributor on https://machengine.org/gpu/ too?

PR #255, opened 2022-04-27
Fix store_op and publicize App.Funcs

silversquirl:
Sorry for not separating this into two PRs, they're very small changes and I couldn't be bothered :)

- I forgot to make App.Funcs public when I created it
- There was a typo in NativeInstance.beginRenderPass that caused `store_op` to use the value of `load_op` instead
---


iddev5:
With the new application interface discussed on matrix and partially implemented with #257, the first change is no longer needed.

slimsag (approved):

slimsag:
Thanks for the fix! I've merged the second change via cherry-pick and dropped the first one per #257

Issue #18, opened 2021-07-30
Browser support


slimsag:
look into how to make this process nice, a unified app/window creation API, etc.

guidorice:
@slimsag  I think it's promising not just for games but for cross-platform graphics of every kind. 👍🏽 

Re: "Browser implementation of gpu.Interface", I am curious if you are planning to use [Emscripten](https://emscripten.org) to build the Javascript <-> WASM glue code? I think that's the only way, currently, because in browser, only Javascript exposes the WebGPU driver.

Also it seems there is definitely interest in using WebGPU for General Purpose GPU (Compute), across native platforms, in browser, and in WASI as well. Some related links, which might provide some ideas for more bespoke things involving shader pipelines etc

* https://tvm.apache.org/2020/05/14/compiling-machine-learning-to-webassembly-and-webgpu
* https://github.com/WebAssembly/WASI/issues/285

PR #257, opened 2022-04-28
mach: Unified entry point and redesigned engine core

iddev5:
From https://matrix.to/#/!hPlvgnLQhMkTucKPrh:matrix.org/$_m_5exLov-5Ue5sGq-6_veTZ0k-RVJAZo9Lj5igjeXA?via=matrix.org

---

This PR implements:
- Unified entry point which would be consistent across all backends (platforms). Right now, we only support desktops which has ``pub fn main`` as their entry point but it doesnt work in other environments like wasm (web) and android or require additional configurations, like in iOS. So instead ask the end application to provide a set of functions: init, deinit, update and resize (and more in future) which are used as per platform limitations behind the scenes
- The bare bones of the ``Engine`` struct which would hold the actual engine functionality. It currently has core (window, events etc) and gpu_driver (webgpu structs). The goal is to provide a cross platform abstraction over these interfaces.

Additionally:
- An ``App`` struct was added in root build.zig which acts as the build system SDK, Its point is to set up the complex build process behind the scenes.
- The ``Options`` which were previously being passed to old ``App`` on init is now managed though an exported declaration in the app file using ``pub const options: mach,Engine.Options = {...};``

This PR does not includes:
- Tests, will need discussion on how to do it, in future (Note that even the old application code didnt had any tests)
- All the examples and shader toy. This is because modifications of examples create ugly diffs and would make this PR difficult to review. In order to not regress the engine, I have ported the examples over in another branch, which I will PR soon keeping really short time difference between both. Or additionally, I can push it over to this branch as soon as review of these below commits are done.

Progress towards #18 

---


slimsag:
I have a few thoughts about the `core` and `gpu_driver` structs, but I can imagine I don't have full picture of how they will be swapped out in a browser context yet so I'm happy to wait to discuss those.

This is a clear improvement over the status quo, so merging!

PR #259, opened 2022-04-30
build: compile examples and applications on demand

iddev5:
Without this change, anytime a dependency is modified, all the examples and applications are rebuilt before tring to run any of them. The problem with this approach is that it leads to long compile times which will keep on increasing as more and more examples are added.

The old behavior is still preserved with a new step ``compile-all``. In future this could be added to CI just to check if all the examples and applications at least build properly. This is not being done right now because zig std lib is undergoing lots of breaking changes right now and it will keep failing CI.

---

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -54,7 +54,7 @@ pub fn build(b: *std.build.Builder) void {
         example_exe.install();
 
         const example_run_cmd = example_exe.run();
-        example_run_cmd.step.dependOn(b.getInstallStep());
+        example_run_cmd.step.dependOn(&example_exe.install_step.?.step);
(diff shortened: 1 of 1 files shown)


slimsag (approved):

slimsag:
> In future this could be added to CI just to check if all the examples and applications at least build properly. This is not being done right now because zig std lib is undergoing lots of breaking changes right now and it will keep failing CI.

We can go ahead and do this now. It's OK if CI fails, that's what it is there for! :) Also, if the CI in this repository fails it won't break the precompiled Dawn builds or anything because those are built in a separate repo's CI.

Plus, our CI uses a pinned version of Zig so it doesn't matter if zig stdlib changes until we try to update to newest version of Zig.

PR #262, opened 2022-05-01
CI: check compilation of all examples

iddev5:
From https://github.com/hexops/mach/pull/259#issuecomment-1113944683

Updates CI to run step ``zig build compile-all`` which compiles all the examples and applications to check if all are at least building properly.

Not added to x86_64-macos because it just fails early with some other error and macos doesnt even have tests enabled, maybe for the same reason?

---


iddev5:
Converted to draft due to #265

slimsag (approved):
LGTM, will probably merge soon.

PR #263, opened 2022-05-01
mach: comptime interface-like implementation of engine cores

iddev5:
Implementation of https://matrix.to/#/!hPlvgnLQhMkTucKPrh:matrix.org/$QYbu5JFyXywFVE-Bdyi2P9HG1nnKBdA0MTkM2u7dtJo?via=matrix.org

---


src/entry_native.zig
```diff
@@ -10,163 +10,6 @@ const c = @import("c.zig").c;
 const Engine = @import("Engine.zig");
 const Options = Engine.Options;
```
slimsag, line 12: Seems like we could maybe put all native code into `native.zig`, instead of having `entry_native.zig` too? not sure what the semantic difference is intended to be

src/Engine.zig
```diff
@@ -63,19 +67,41 @@ delta_time_ns: u64 = 0,
...
+
+    return engine;
+}
+
+fn GetCoreInternalType() type {
+    return @import("native.zig").CoreGlfw;
+}
+
+fn GetGpuDriverInternalType() type {
+    return @import("native.zig").GpuDriverNative;
```
slimsag, line 106: I haven't looked closely enough yet to see how viable this is, but here's what I am thinking:

* Maybe instead of two separate `Core` and `GpuDriver` types, we could have just one `Platform` type (which handles platform specifics)
* Maybe we should not use the name `internal` (implies "never touch this") and instead use `platform` (it would be possible to expose platform-specific functionality behind these, like say ability to send a push notification on android/ios as an example - although I don't know we actually would want such a feature.)

Just a thought I'll explore later, sharing in case any of it resonates with you. Happy to continue down this code path / design for now 👍

slimsag (approved):
Great work again!

src/entry_native.zig
iddev5: I don't really like entry_native.zig, so sounds like a good plan to me. The whole organization is still a bit experimental I would say

src/Engine.zig
iddev5: 1) Yeah that would work well especially because Core and GpuDriver are quite related 
2) Agreed

PR #269, opened 2022-05-08
examples: fixed uniform buffer release and unused var

PiergiorgioZagaria:

### DIFF
--- a/examples/instanced-cube/main.zig
+++ b/examples/instanced-cube/main.zig
@@ -122,7 +122,6 @@ pub fn init(app: *App, engine: *mach.Engine) !void {
         .size = @sizeOf(UniformBufferObject) * num_instances,
         .mapped_at_creation = false,
     });
-    defer uniform_buffer.release();
     const bind_group = engine.gpu_driver.device.createBindGroup(
         &gpu.BindGroup.Descriptor{
             .layout = bgl,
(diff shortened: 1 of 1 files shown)


slimsag (approved):

examples/instanced-cube/main.zig
```diff
@@ -122,7 +122,6 @@ pub fn init(app: *App, engine: *mach.Engine) !void {
         .size = @sizeOf(UniformBufferObject) * num_instances,
         .mapped_at_creation = false,
     });
-    defer uniform_buffer.release();
     const bind_group = engine.gpu_driver.device.createBindGroup(
         &gpu.BindGroup.Descriptor{
             .layout = bgl,
```
slimsag, line 127: I think the reason this worked and was not UB, is because `createBindGroup` takes this `uniform_buffer` as a parameter. So it probably `.reference()`'s the `uniform_buffer` parameter and holds on to it. I think we maybe maintaining a duplicate `reference` needlessly now.

But I need to confirm this / dig into the Dawn source code & make sure this is true, review all examples for correctness, etc.

PiergiorgioZagaria: I think that `release()` just puts the memory that was allocated before in free memory, but doesn't change neither the contents neither uniform_buffer that points to what is now freed memory. Hence we would have probably seen something strange if we later allocated and wrote something to new allocated memory.

PR #270, opened 2022-05-08
examples: created gkurve example

PiergiorgioZagaria:

There is still much missing, like textures, correct matrices for positioning and some form of dynamic arrays in wgsl, but it's a start

### DIFF
--- a/examples/gkurve/frag.wgsl
+++ b/examples/gkurve/frag.wgsl
@@ -1,0 +1,123 @@
+//! Ported from https://www.shadertoy.com/view/ltXSDB
+
+// Signed Distance to a Quadratic Bezier Curve
+// - Adam Simmons (@adamjsimmons) 2015
+//
+// License Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported License
+//
+// Inspired by http://www.pouet.net/topic.php?which=9119
+// and various shaders by iq, T21, and demofox
+// 
+// I needed the -signed- distance to a quadratic bezier
+// curve but couldn't find any examples online that
+// were both fast and precise. This is my solution.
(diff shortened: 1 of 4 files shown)


examples/gkurve/frag.wgsl
```diff
@@ -0,0 +1,123 @@
...
+    dis = min(dis, length(pos - p));
+    
+    return dis * signBezier(A, B, C, p);
+}
+
+
+@stage(fragment) fn main( 
+    @location(0) uv : vec2<f32>,
+    @interpolate(flat) @location(1) instance_index: u32,
+) -> @location(0) vec4<f32> {
```
slimsag, line 88: Interesting how (relatively) complex this frag shader is. If my reading is correct, it's calculating barycentric coordinates for each fragment in the shader.

I think we can simplify this a lot (at least in terms of code, maybe with better perf too) by passing the barycentric coordinates of the triangle vertices as attributes to the vertex shade (simply `(1, 0, 0)`, `(0, 1, 0)`, and `(0, 0, 1)` for the 3 vertices of the triangle, as described e.g. [here](https://stackoverflow.com/a/25415763)) and then have the vertex shader pass them to the fragment shader as varying - doing the interpolation for us so we have barycentric coordinates per fragment - and then use the rather simple loop & blinn shader to discard fragments on the side we don't want as e.g. [described here](https://www.reddit.com/r/opengl/comments/1a1vai/loop_and_blinn_quadratic_curves/c8tk9fi/)

slimsag (approved):
nice work, great starting point!

I'll look into if I can simplify this a bit. let's keeping discussing in the Matrix chat to see where we can bring this

PR #273, opened 2022-05-12
mach: Create binding methods for all glfw methods in use and update examples

iddev5:
Implemented the rest of raw glfw functions which were still in use. The added methods follow the same glfw naming, but can be changed later if desired.

Updated the examples with related to same.

---


slimsag (approved):
Nice work!

PR #279, opened 2022-05-17
mach: introduce cross platform Timer abstraction

iddev5:
This Timer uses std.time.Timer as backing timer in native platforms, and will use custom timers for special platforms (wasm, android?, ios?).

Unlike std.time.Timer, its primary API is focused on floats. Also meant to provides some convenient functions alongside base ones.

Follows std.time.Timer API, but methods by default return f32 i.e non-precise variant with precise variants available returning u64.

### DIFF
--- a/src/Timer.zig
+++ b/src/Timer.zig
@@ -1,0 +1,41 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
+const Timer = @This();
+
+backing_timer: BackingTimerType = undefined,
+
+// TODO: verify declarations and its signatures
+const BackingTimerType = if (builtin.cpu.arch == .wasm32) void else std.time.Timer;
+
+/// Initialize the timer.
+pub fn start() !Timer {
+    return Timer{
+        .backing_timer = try BackingTimerType.start(),
+    };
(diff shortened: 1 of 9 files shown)


src/Timer.zig
```diff
@@ -0,0 +1,41 @@
...
+    timer.backing_timer.reset();
+}
+
+/// Returns the current value of the timer in nanoseconds, then resets it.
+pub fn lapPrecise(timer: *Timer) u64 {
+    return timer.backing_timer.lap();
+}
+
+/// Returns the current value of the timer in seconds, then resets it.
+pub fn lap(timer: *Timer) f32 {
```
slimsag, line 39: BTW, probably we should mark all "wrapper" functions like this with `inline` just to be on the safe side?

slimsag (approved):

src/Timer.zig
iddev5: Alright, I will revisit it soon.

InKryption: Seems like a wise decision; probably shouldn't leave it up to the inline quota.

PR #282, opened 2022-05-20
mach: Event loop based input handling

iddev5:
Relevent commit message:
```
This commit changes the former callback based design to handle key input
(GLFW-like) to an event loop based design (SDL-like). This uses a
TailQueue to store the events from inside of standard glfw callbacks.
This Queue is then popped while polling, thereby emulating event loop.

Removes from Engine the function: ``setKeyCallback`` and adds the
function: ``pollEvent`` which may return an event or null.

This change was done for two reasons:
1) Removing dependence of Engine on App. This was a circular dependency
   and a genuine bad design.
2) Solve the recent regression due to the same which was (i) preventing
   using types declared in Engine.zig and (ii) preventing usage of
   multiple source files in an application.

Currently only key press and release events are implemented as these are
the ones currently used in examples.
```

---

### DIFF
--- a/examples/advanced-gen-texture-light/main.zig
+++ b/examples/advanced-gen-texture-light/main.zig
@@ -33,7 +33,6 @@ const Dir = struct {
 };
 
 pub fn init(app: *App, engine: *mach.Engine) !void {
-    engine.core.setKeyCallback(keyCallback);
     // todo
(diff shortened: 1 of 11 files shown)


slimsag:
Force-pushed a rebase to fix conflict after I merged https://github.com/hexops/mach/pull/280

PR #284, opened 2022-05-22
examples/gkurve: added shapes and textures

PiergiorgioZagaria:


slimsag (approved):
Looks pretty great to me. Random thoughts:

* We definitely need a better way to do the circles, I'm going to investigate this. I think we can add another type of primitive which approximates a circle.
* Our orthographic view is currently in subpixel units, so everything is 2x smaller on macOS than on Windows/Linux machines. I'll fix this so our view is always in pixel (not subpixel) units.

Looking forward to the texture atlas approach as well!

Issue #232, opened 2022-04-18
build: Don't install libraries


silversquirl:
Currently, all libraries produced by Mach are installed to the output prefix. This is undesirable as Zig's build system can be used to produce packages, or to install globally to the system, and Mach static libs are simply compilation side-products, not necessarily things wanted in the installation.

If building Dawn for non-Zig use is a supported usecase, I would suggest having a separate API for this - the current one doesn't really support it well anyway.

On a related note, `mach/gpu/build.zig` and `mach/build.zig` also build static libraries for some reason? Not sure why that is; they probably shouldn't since they're just Zig packages and don't expose C APIs.

Luexa:
> Currently, all libraries produced by Mach are installed to the output prefix. This is undesirable as Zig's build system can be used to produce packages, or to install globally to the system, and Mach static libs are simply compilation side-products, not necessarily things wanted in the installation.

Oh yeah, I meant to post something about this but didn't have the time earlier. Installing should never be done automatically; instead the `LibExeObjStep` should be exposed somewhere so the user can invoke `.install()` if they wish.

> If building Dawn for non-Zig use is a supported usecase, I would suggest having a separate API for this - the current one doesn't really support it well anyway.

Yup, there is no canonical way to link to the libraries as these libraries are local variables that are never returned. This, in turn, exacerbates the problem. As I said above the `LibExeObjStep` should be exposed if these libraries are intended for user to link to (if they are not, they should not be generated in the first place, let alone installed).

> On a related note, `mach/gpu/build.zig` and `mach/build.zig` also build static libraries for some reason? Not sure why that is; they probably shouldn't since they're just Zig packages and don't expose C APIs.

Yeah this was confusing for me as well. It essentially creates a dud library that further pollutes the install root.

slimsag:
Thanks for filing this, and for reaching out earlier about the same issue @Luexa !

I removed the unused static libraries from `mach/gpu/build.zig` and `mach/build.zig` in #220 

> the `LibExeObjStep` should be exposed somewhere so the user can invoke `.install()` if they wish.

Makes sense to me, thanks for that idea!

> If building Dawn for non-Zig use is a supported usecase, I would suggest having a separate API for this - the current one doesn't really support it well anyway.

[...]

PR #286, opened 2022-05-22
mach: initial wasm platform implementation

iddev5:
This is very incomplete, it is not ready to support the examples yet.
This PR just implements the barebones structures needed for future wasm development. The build system support is also lacking, which will be done progressively.

In order to run:
```sh-session 
# This is actually going to compile all examples but you can only run the last one
zig build compile-all -Dtarget=wasm32-freestanding-none 
cd zig-out/lib
cp ../../www/template.html application.html
cp ../../src/mach.js .
# Run a web server
python3 -m http.server
# Navigate to https://localhost:8000/application.html
```
This PR also implements some general build.zig sdk improvements:
```
- Target is now passed as an option in App.init()
- Always install() in all case
- Add functions getInstallStep(), setBuildMode() and run()
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -59,6 +59,7 @@ pub fn build(b: *std.build.Builder) void {
         );
         example_app.setBuildMode(mode);
         example_app.link(options);
+        example_app.install();
 
         if (target.toTarget().cpu.arch != .wasm32) {
             const example_run_cmd = example_app.run();
@@ -80,6 +81,7 @@ pub fn build(b: *std.build.Builder) void {
         );
         shaderexp_app.setBuildMode(mode);
         shaderexp_app.link(options);
+        shaderexp_app.install();
 
         const shaderexp_run_cmd = shaderexp_app.run();
         shaderexp_run_cmd.step.dependOn(&shaderexp_app.getInstallStep().?.step);
@@ -149,13 +151,16 @@ const App = struct {
 
         step.addPackage(app_pkg);
         step.setTarget(options.target);
-        step.install();
 
         return .{
             .b = b,
             .step = step,
             .name = options.name,
         };
+    }
+
+    pub fn install(app: *const App) void {
+        app.step.install();
     }
 
     pub fn link(app: *const App, options: Options) void {
--- a/build.zig
+++ b/build.zig
@@ -31,15 +31,23 @@ pub fn build(b: *std.build.Builder) void {
 
     inline for ([_]ExampleDefinition{
         .{ .name = "triangle" },
-        //.{ .name = "boids" },
+        .{ .name = "boids" },
         .{ .name = "rotating-cube", .packages = &[_]Pkg{Packages.zmath} },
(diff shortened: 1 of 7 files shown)


slimsag:
> Always install() in all case

What is the motivation behind this? Are we sure it's always the right default? Especially given conversations like #232

iddev5:
I did that under the assumption that applications are always installed unconditionally. This can be changed though, its not an important thing to have.

iddev5:
Well i undid this, there are no clear benefits.

build.zig
```diff
@@ -35,49 +35,59 @@ pub fn build(b: *std.build.Builder) void {
...
+        .{ .name = "gkurve", .packages = &[_]Pkg{ Packages.zmath, Packages.zigimg }, .std_platform_only = true },
+        .{ .name = "textured-cube", .packages = &[_]Pkg{ Packages.zmath, Packages.zigimg }, .std_platform_only = true },
     }) |example| {
+        // FIXME: this is workaround for a problem that some examples (having the std_platform_only=true field) as
+        // well as zigimg uses IO which is not supported in freestanding environments. So break out of this loop
+        // as soon as any such examples is found. This does means that any example which works on wasm should be
+        // placed before those who dont.
+        if (example.std_platform_only)
+            if (target.toTarget().cpu.arch == .wasm32)
+                break;
```
slimsag, line 49: Would a `continue` (skip this iteration) not work here?

iddev5: No, it doesnt work. Control flows inside inline for is quite broken.

src/mach.js
```diff
@@ -0,0 +1,90 @@
...
+      original_title;
+
+    mach.canvases[canvas].title = str;
+  },
+
+  machCanvasSetSize(canvas, width, height) {
+    const cv = mach.canvases[canvas];
+    if (width > 0 && height > 0) {
+      cv.canvas.width = width;
+      cv.canvas.height = height;
```
slimsag, line 75: I think according to https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio we should use:

```
cv.canvas.width = Math.floor(width * window.devicePixelRatio);
cv.canvas.height = Math.floor(height * window.devicePixelRatio);
cv.canvas.style.width = width + "px";
cv.canvas.style.height = height + "px";
```

This way we get a canvas size that has sub-pixel resolution on e.g. Retina displays

src/mach.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+  },
+
+  machCanvasGetWidth(canvas) {
```
slimsag, line 79: Should change these to return `Math.floor(cv.canvas.width / window.devicePixelRatio)` I think, so it returns e.g. `640` instead of `1280` on Retina displays

Then introduce a `machCanvasFramebufferGetWidth` which returns `cv.canvas.width` (1280 not 640)

slimsag (approved):
This looks great! It's *REALLY* exciting to see this start to come together!

PR #289, opened 2022-05-23
gpu: throw error if texture data is not slice

alichraghi:

### DIFF
--- a/gpu/src/Queue.zig
+++ b/gpu/src/Queue.zig
@@ -63,9 +63,10 @@ pub inline fn writeBuffer(queue: Queue, buffer: Buffer, buffer_offset: u64, comp
 pub inline fn writeTexture(
     queue: Queue,
     destination: *const ImageCopyTexture,
-    data: anytype,
     data_layout: *const Texture.DataLayout,
     write_size: *const Extent3D,
+    comptime T: type,
+    data: []const T,
 ) void {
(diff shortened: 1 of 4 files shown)


gpu/src/Queue.zig
```diff
@@ -67,6 +67,9 @@ pub inline fn writeTexture(
     data_layout: *const Texture.DataLayout,
     write_size: *const Extent3D,
 ) void {
+    if (@TypeOf(data) != []const u8) {
```
slimsag, line 70: Nice! I think there are more places than just here that we need to update - though.

* `Queue.zig` `fn writeBuffer`
* `CommandEncoder.zig` `fn writeBuffer`

alichraghi: yeah im aware. will update this soon

alichraghi: also this seems wrong (check CI log)

slimsag: It looks like `std.meta.isSliceOf(T)` might do what we want https://sourcegraph.com/github.com/ziglang/zig/-/blob/lib/std/meta/trait.zig?L142:7

alichraghi: yeah make sense :+1:

slimsag (approved):
Even better, thank you for doing this!

PR #302, opened 2022-05-25
wasm: implement more functionality

iddev5:
Apart from general improvements, it implements:
- https://github.com/hexops/mach/pull/286#discussion_r878909929
- Key events

Simple program I prepared for testing: https://gist.github.com/iddev5/ddbd9290e0a55b27fd0af5cbefb65aff
The process to run is the same as mentioned in #286 except that add the example to build.zig first. The proper build system run support will come in a different PR.

---

### DIFF
--- a/src/mach.js
+++ b/src/mach.js
@@ -42,8 +42,10 @@ const mach = {
   machCanvasInit(width, height, id) {
     let canvas = document.createElement("canvas");
     canvas.id = "#mach-canvas-" + mach.canvases.length;
-    canvas.width = width;
-    canvas.height = height;
+    canvas.style.width = width + "px";
+    canvas.style.height = height + "px";
+    canvas.width = Math.floor(width * window.devicePixelRatio);
+    canvas.height = Math.floor(height * window.devicePixelRatio);
     canvas.tabIndex = 1;
 
     mach.setString(canvas.id, id);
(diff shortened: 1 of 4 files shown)


src/mach.js
```diff
@@ -100,6 +239,13 @@ const mach = {
     return cv.canvas.height;
   },
 
+  machEventShift() {
+    if (mach.events.length < 0)
+      return 0;
+
+    return mach.events.shift();
```
iddev5, line 246: Note to self: Possible optimization: Use an array and a cursor indicating the current head of the list. Each time we call this function, return list[cursor] and then increment it. Finally reset the list and cursor at the end of frame (needs a new function)

The point is that shift operation may involve moving of all array elements which is performance intensive. In practice, it is likely that JS engines optimize this properly.

src/mach.js
```diff
@@ -71,21 +212,40 @@ const mach = {
...
+  machCanvasGetFramebufferHeight(canvas) {
     const cv = mach.canvases[canvas];
     return cv.canvas.height;
   },
 
+  machEventShift() {
+    if (mach.events.length < 0)
+      return 0;
+
+    return mach.events.shift();
```
iddev5, line 246: Note to self: Possible optimization: Use an array and a cursor indicating the current head of the list. Each time we call this function, return list[cursor] and then increment it. Finally reset the list and cursor at the end of frame (needs a new function)

The point is that shift operation may involve moving of all array elements which is performance intensive. In practice, it is likely that JS engines optimize this properly.

slimsag (approved):
This looks awesome!

PR #306, opened 2022-05-26
mach: fix build.zig

d3m1gd:
fixes to allow using mach as a library in standalone projects


iddev5 (changes requested):
I think you have accidentally updated zigimg with this PR

slimsag (approved):

PR #310, opened 2022-05-26
examples/gkurve: added atlas text rendering and freetype linking

PiergiorgioZagaria:


PiergiorgioZagaria:
It's still very WIP, I think I will start working on the scalable text now. I will move text.zig to something like label.zig and create another scalable_label.zig file with the other implementation. After that I will work on a major refactor of the whole draw API.

slimsag (approved):

slimsag:
nice work!

One thing I notice is we have bleeding between the glyphs (right side of letters B and S):

image

Probably we need to add more margin between items in the texture atlas.

slimsag:
(I fixed the merge conflict and made the build.zig change simpler)

PR #313, opened 2022-05-27
Mach: improvements to build system (for wasm)

iddev5:
- Made general improvements to build system on how to add dependencies, removed redundant dependencies internally
- In case of wasm, install the necessary js and html files in prefix/www. The main HTML file is generated with an application placed in tools/html-generator.zig. This application would prove to be more useful in future when we have more modules inside the engine (as all js files need to be listen inside this HTML file)
- In case of wasm, serve the application using a http server placed in tools/http-server.zig. The address and port can be customized using the environment variables ``MACH_ADDRESS`` and ``MACH_PORT`` respectively.
- Finally in case of wasm, launch a web browser with the link where the application is being served.

The serve + launch steps can be customized with options in future, but right now it's kept as simple as possible.

This PR enables use of run step for wasm application. Now you can freely do ``zig build run-example-* -Dtarget=wasm32-freestanding-none`` and everything will just work (regarding the build system)

Further, now generate separate html and wasm files in install:
```
# ls zig-out/www
example-triangle.html  example-triangle.wasm  example-two-cubes.html  example-two-cubes.wasm  main.js ...
```

With this PR we achieve feature parity for general support of wasm development.

Fixes hexops/mach#18

---

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "freetype/upstream"]
 	path = freetype/upstream
 	url = https://github.com/hexops/freetype
+[submodule "tools/libs/apple_pie"]
+	path = tools/libs/apple_pie
+	url = https://github.com/Luukdegram/apple_pie
--- a/build.zig
+++ b/build.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const builtin = @import("builtin");
 pub const gpu = @import("gpu/build.zig");
 const gpu_dawn = @import("gpu-dawn/build.zig");
 pub const glfw = @import("glfw/build.zig");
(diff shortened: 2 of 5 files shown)


GithubPrankster:
Successfully works in Windows+Chromium and Windows+Firefox. Firefox may require serving html files as `index.html` (issue encountered in Librewolf fork)

slimsag:
I made a small change to ensure the apple_pie submodule is cloned automatically

slimsag (approved):

slimsag:
Damn, this is so cool @iddev5!

I can also confirm this works on macOS+Firefox and macOS+Safari

PR #316, opened 2022-05-29
glfw: ensure git present

d3m1gd:

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -25,6 +25,7 @@ pub fn build(b: *std.build.Builder) void {
     test_step.dependOn(&main_tests.step);
 
     // TODO(build-system): https://github.com/hexops/mach/issues/229#issuecomment-1100958939
+    ensureGit(b.allocator);
     ensureDependencySubmodule(b.allocator, "examples/libs/zmath") catch unreachable;
     ensureDependencySubmodule(b.allocator, "examples/libs/zigimg") catch unreachable;
     ensureDependencySubmodule(b.allocator, "examples/assets") catch unreachable;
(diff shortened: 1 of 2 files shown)


iddev5 (commented):
Top level build.zig also needs this check as it contains ensureSubmodule()

d3m1gd:
it might not be needed there, as glfw will be called anyaway, and missing git will be discovered

iddev5:
That won't work. Non-native (wasm) builds don't call glfw.link().

build.zig
```diff
@@ -211,10 +211,31 @@ fn ensureDependencySubmodule(allocator: std.mem.Allocator, path: []const u8) !vo
     if (std.process.getEnvVarOwned(allocator, "NO_ENSURE_SUBMODULES")) |no_ensure_submodules| {
         if (std.mem.eql(u8, no_ensure_submodules, "true")) return;
     } else |_| {}
+    ensureGit(allocator);
```
iddev5, line 214: I think this should be rather put here: https://github.com/hexops/mach/blob/0dabe0d49fa621ed43dc2ba092f56e54d4b117b3/build.zig#L26
There's no need to call this function everytime you invoke git

glfw/system_sdk.zig
```diff
@@ -216,6 +216,7 @@ fn determineSdkRoot(allocator: std.mem.Allocator, org: []const u8, name: []const
         return sdk_root_dir;
     } else |err| return switch (err) {
         error.FileNotFound => {
+            ensureGit(allocator);
```
iddev5, line 219: And this should be placed here: https://github.com/hexops/mach/blob/f0f91874da865d43eabb751905dd8e0f9026f09e/glfw/system_sdk.zig#L204 as both the code paths use git

slimsag: agreed

d3m1gd:
feel free to squash as needed

iddev5 (approved):

slimsag (approved):

PR #320, opened 2022-05-30
mach: save window and framebuffer size from glfw callbacks

david-vanderson:
Previously if user code called core.getWindowSize() (or
getFramebufferSize) during their update they could get sizes that did
not match engine.gpu_driver.target_desc because the window had changed
size.

Now core.getWindowSize() returns the last size we got from a glfw
callback instead of directly querying the window.  So all sizes should
agree during a single frame.

This relies on glfw callbacks only happening when we call
glfw.pollEvents().  This isn't strictly true - a callback could happen
during any glfw call.  Might have to revisit this later.

### DIFF
--- a/src/platform/native.zig
+++ b/src/platform/native.zig
@@ -15,6 +15,9 @@ pub const Core = struct {
     events: EventQueue = .{},
     user_ptr: UserPtr = undefined,
 
+    last_window_size: structs.Size,
+    last_framebuffer_size: structs.Size,
+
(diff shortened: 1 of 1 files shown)


slimsag (approved):
Nice work!

> This relies on glfw callbacks only happening when we call
glfw.pollEvents(). This isn't strictly true - a callback could happen
during any glfw call. Might have to revisit this later.

I would be very surprised if this is the case. I don't think it is.

Issue #292, opened 2022-05-23
gpu-dawn binaries for Mac not available for latest zig


kamidev:
Zig 0.10 has changed the Darwin default ABI from "gnu" to "none", with this breaking PR: https://github.com/ziglang/zig/pull/11684. As a consequence, the expected gpu-dawn binaries currently does not exist for Mac users. 

Below an example using zig-gamedev. The problem is general, though.

```
➜  zig-gamedev git:(main) rm -rf git-cache; stage1 build
error: gpu-dawn binaries for aarch64-macos.12.4...12.4-none not available.
error: -> open an issue: https://github.com/hexops/mach/issues
error: -> build from source (takes 5-15 minutes):
error:        use -Ddawn-from-source=true or set `Options.from_source = true`
...
```
Building from source still works as expected.

slimsag:
Thanks for filing, we're not yet up-to-date with that Zig version. We generally try to track latest nightly, but not always up to date. You can find the latest supported version Mach tracks in the CI config here:

https://github.com/hexops/mach/blob/f3e2b8b74c38f829e5717a75f6a07017ab6b8872/.github/workflows/ci.yml#L94

I'll leave this issue open until we do update.

slimsag:
This is fixed, and you can now find the version of Zig we test with here: https://github.com/hexops/mach#supported-zig-version

Should be easier for us to keep up to date in the future, too.

Let me know if you run into any trouble.

kamidev:
Works like a charm! We just changed zig-gamedev to use the right Mac binaries. Thanks a lot.

PR #321, opened 2022-05-31
CI: automate Zig version upgrades; upgrade Zig & more

slimsag:
after this change, it should be possible for anyone to easily send a PR to update the Zig version we use:

- To upgrade Zig version, we just run e.g. `ZIG_VERSION='zig-0.10.0-dev.2431+0e6285c8f' ./dev/update-zig.sh` and it will update every CI yml file in the repository, as well as the README.
- The README now states the currently-tested Zig version.
- The M1 runner now has it's Zig version managed by the yml file (previously me manually installing it on the machine.)

### DIFF
--- a/.github/workflows/m1_ci.yml
+++ b/.github/workflows/m1_ci.yml
@@ -1,4 +1,4 @@
-name: M1 CI
+name: CI
 on:
   # SECURITY: This must be a push event only, otherwise our M1 mac runner would be compromised by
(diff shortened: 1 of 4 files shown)


michal-z:
Nice! Does this also fix binary versions for macos?

slimsag:
It's going to - but it's not yet fixed. I'll close https://github.com/hexops/mach/issues/292 once fixed.

PR #323, opened 2022-05-31
README: Fix dead link to article

iddev5:


slimsag:
Ahg, thanks for sending this - the link should actually work, since it's also in other places (HN, Reddit, etc.). I guess the migration to Hugo broke it.

Fixed in https://github.com/hexops/devlog/commit/7256d888994a46e58341cdf9f221ee70854b3a45 just now.

PR #325, opened 2022-06-03
add mouse motion and mouse button events

david-vanderson:

### DIFF
--- a/src/structs.zig
+++ b/src/structs.zig
@@ -56,9 +56,11 @@ pub const Event = union(enum) {
         x: f64,
         y: f64,
     },
-    mouse_button: struct {
+    mouse_press: struct {
         button: enums.MouseButton,
-        action: enums.Action,
+    },
+    mouse_release: struct {
+        button: enums.MouseButton,
     },
     scroll: struct {
         xoffset: f64,
--- a/src/structs.zig
+++ b/src/structs.zig
@@ -51,4 +51,13 @@ pub const Event = union(enum) {
     key_release: struct {
         key: enums.Key,
     },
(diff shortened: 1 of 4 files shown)


src/structs.zig
```diff
@@ -51,4 +51,13 @@ pub const Event = union(enum) {
...
         key: enums.Key,
     },
+    mouse_motion: struct {
+        // These are in window coordinates (not framebuffer coords)
+        x: f64,
+        y: f64,
+    },
+    mouse_button: struct {
+        button: enums.MouseButton,
+        action: enums.Action,
```
iddev5, line 61: I m wondering if using Action is the right thing to do versus separate events for press and release (button_press, button_release) like we have for key events. The key events has been named so because many libraries choose to differentiate the events completely instead of having one event type for the category.

iddev5 (approved):

src/structs.zig
david-vanderson: Yes I'm also unsure.  In my gui library I started with mouse events being separated (leftdown, leftup, rightdown, rightup, etc.) and am moving back towards the glfw model with separate Action.  It seems to align the event handlers with how I think of them ("I'm handling mouse events here, then split by press/release).

But I don't think it matters that much.  I'll switch it.  Thanks!

david-vanderson: Also it matters a lot more that we are consistent.

iddev5:
I think mouse_press and mouse_release is better from an API point of view. When one calls pollEvent() they expect to get an event. mouse_button is not an event, but mouse_press and mouse_release are.

PR #326, opened 2022-06-03
add hasEvent(), waitEvent(), and waitEventTimeout()

david-vanderson:

### DIFF
--- a/src/Engine.zig
+++ b/src/Engine.zig
@@ -58,10 +58,10 @@ pub fn setShouldClose(engine: *Engine, value: bool) void {
     engine.internal.setShouldClose(value);
 }
 
-// Signals mach to wait for an event with timeout before calling update()
-// again.  Mach resets to null, so call during each update() if needed.
+// Sets seconds to wait for an event with timeout before calling update()
+// again.
 //
-// timeout is in seconds (null disables waiting)
+// timeout is in seconds (<= 0.0 disables waiting)
 // - pass std.math.floatMax(f64) to wait with no timeout
 //
 // update() can be called earlier than timeout if an event happens (key press,
@@ -69,9 +69,7 @@ pub fn setShouldClose(engine: *Engine, value: bool) void {
 //
 // update() can be called a bit later than timeout due to timer precision and
 // process scheduling.
-//
-// Calling this multiple times means the last call's value is used.
-pub fn setWaitEvent(engine: *Engine, timeout: ?f64) void {
+pub fn setWaitEvent(engine: *Engine, timeout: f64) void {
     engine.internal.setWaitEvent(timeout);
 }
 
@@ -89,7 +87,7 @@ pub fn getWindowSize(engine: *Engine) structs.Size {
     return engine.internal.getWindowSize();
 }
 
-pub fn hasEvent(engine: *Engine) !bool {
+pub fn hasEvent(engine: *Engine) bool {
     return engine.internal.hasEvent();
 }
 
--- a/src/Engine.zig
+++ b/src/Engine.zig
@@ -53,8 +53,26 @@ pub fn setOptions(engine: *Engine, options: structs.Options) !void {
     engine.options = options;
 }
 
+// Signals mach to stop the update loop.
 pub fn setShouldClose(engine: *Engine, value: bool) void {
     engine.internal.setShouldClose(value);
+}
+
+// Signals mach to wait for an event with timeout before calling update()
+// again.  Mach resets to null, so call during each update() if needed.
+//
+// timeout is in seconds (null disables waiting)
+// - pass std.math.floatMax(f64) to wait with no timeout
+//
+// update() can be called earlier than timeout if an event happens (key press,
+// mouse motion, etc.)
+//
+// update() can be called a bit later than timeout due to timer precision and
+// process scheduling.
+//
+// Calling this multiple times means the last call's value is used.
+pub fn setWaitEvent(engine: *Engine, timeout: ?f64) void {
+    engine.internal.setWaitEvent(timeout);
 }
 
 // Returns the framebuffer size, in subpixel units.
@@ -69,6 +87,10 @@ pub fn getFramebufferSize(engine: *Engine) structs.Size {
 // e.g. returns 1280x960 on macOS for a window that is 640x480
 pub fn getWindowSize(engine: *Engine) structs.Size {
     return engine.internal.getWindowSize();
+}
+
+pub fn hasEvent(engine: *Engine) !bool {
+    return engine.internal.hasEvent();
 }
 
 pub fn pollEvent(engine: *Engine) ?structs.Event {
--- a/src/platform/native.zig
+++ b/src/platform/native.zig
@@ -17,7 +17,7 @@ pub const Platform = struct {
 
     last_window_size: structs.Size,
     last_framebuffer_size: structs.Size,
-    waitEventTimeout: ?f64 = null,
+    wait_event_timeout: f64 = 0.0,
 
     native_instance: gpu.NativeInstance,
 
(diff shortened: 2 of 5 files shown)


src/platform/native.zig
```diff
@@ -256,6 +256,21 @@ pub const Platform = struct {
...
+    }
+
+    pub fn waitEvent(platform: *Platform) !void {
+        _ = platform;
+        try glfw.waitEvents();
+    }
+
+    pub fn waitEventTimeout(platform: *Platform, timeout: f64) !void {
+        _ = platform;
+        try glfw.waitEventsTimeout(timeout);
```
slimsag, line 271: Would like to hear @iddev5's thoughts on this PR - I am curious about feasability of implementing in browser.

david-vanderson: Me too!  I guess if needed it could be implemented by doing short sleeps and checking for events between.

iddev5: waitEvent in general could not be implemented in web. There can be workaround like a canUpdate() function toogled by events which will prevent update() from being called in js.

david-vanderson: I'm trying to understand how the js version of mach works.  How does wasmUpdate() get called?

My mental model is that it would be called either in response to an input event or a timer.  If that's the case, then waitEvent could return control to the browser and disable the timer.  Can you help me understand?

I'd love to try and get the wasm backend running so I can help there too.  What's the zig build command that you use?

david-vanderson: Thinking some more, maybe an alternative would be to return info from update() that told the engine whether to wait for an event or to call update() again as soon as possible.  The return type of update() could be !?f64 where null means "call again no waiting" and f64 means "wait for an event with this timeout in seconds"?

iddev5: Js (browsers) can't run an infinite loop because a webpage itself is running with an init-update-deinit loop. So the alternative they provide is a function called ``requestAnimationFrame``. You pass in a callback to this function which appends it to an internal list. At some point in the browser's internal update loop, it pops and executes all the function. This is how update is called for one frame. In order to repeat it for multiple frame, you requestAnimationFrame you update function again as soon as the first one is done. Pseudo code:
```
function update() {
    wasmUpdate();
    requestAnimationFrame(update);
}
requestAnimationFrame(update);
``` 
(Its not recursion)

The code is here in mach: https://github.com/hexops/mach/blob/main/www/template.html#L27-L33

---

Run a wasm application using ``zig build run-example-* -Dtarget=wasm32-freestanding-none``, but the slight problem is that since we dont support webgpu, we get weird integer truncation errors from mach-gpu. So the solution is to make a custom simple example which doesnt do any gpu calls.

david-vanderson: Thanks very much for the insight!  I'm digging into it.

david-vanderson:
This is at least the native side of it.

src/Engine.zig
```diff
@@ -53,10 +53,28 @@ pub fn setOptions(engine: *Engine, options: structs.Options) !void {
     engine.options = options;
 }
 
+// Signals mach to stop the update loop.
 pub fn setShouldClose(engine: *Engine, value: bool) void {
     engine.internal.setShouldClose(value);
 }
 
+// Signals mach to wait for an event with timeout before calling update()
+// again.  Mach resets to null, so call during each update() if needed.
```
iddev5, line 62: Why are we resetting it every frame? My general understanding states that setXyz() functions are meant to do permanent changes.

src/Engine.zig
```diff
@@ -53,10 +53,28 @@ pub fn setOptions(engine: *Engine, options: structs.Options) !void {
...
+//
+// timeout is in seconds (null disables waiting)
+// - pass std.math.floatMax(f64) to wait with no timeout
```
iddev5, line 65: Why are we using ``std.math.floatMax(f64)`` instead of just 0 (zero)?

src/platform/native.zig
```diff
@@ -17,6 +17,7 @@ pub const Platform = struct {
 
     last_window_size: structs.Size,
     last_framebuffer_size: structs.Size,
+    waitEventTimeout: ?f64 = null,
```
iddev5, line 20: Nit: variable case

src/platform/native.zig
```diff
@@ -305,6 +306,15 @@ pub const Platform = struct {
         return platform.last_window_size;
     }
 
+    pub fn hasEvent(platform: *Platform) !bool {
+        try glfw.pollEvents();
```
iddev5, line 310: I dont think hasEvents should poll, we already do that here: https://github.com/hexops/mach/blob/main/src/platform/native.zig#L488

src/platform/native.zig
```diff
@@ -512,5 +522,19 @@ pub fn main() !void {
         }
 
         try app.update(&engine);
+
+        if (engine.internal.waitEventTimeout) |timeout| {
```
iddev5, line 526: I dont think this is the correct usage. From glfw docs, waitEvents works exactly like pollEvents except that it blocks until there is one or more events. So this whole section should be placed in if-else of the glfw.pollEvents() call above.

iddev5 (changes requested):

src/Engine.zig
david-vanderson: Hadn't thought of making the change permanent.  I think most people will either not call it at all or call it every frame.  I can make that change.

src/Engine.zig
david-vanderson: I'm thinking of changing the timeout from ?f64 to f64, where <= 0 disables waiting (which is the same as "wait 0 seconds").  Then there are no special cases - if you want an effectively infinite timeout, use floatMax(f64).  That's the reasoning.

src/platform/native.zig
david-vanderson: will fix

src/platform/native.zig
david-vanderson: Thanks for making me rethink this.  You are right I can do without the poll.

src/platform/native.zig
david-vanderson: I think I understand, let me try.

david-vanderson:
Thanks for the review!  How does it look now?

src/Engine.zig
iddev5: That could work, I think.

iddev5 (approved):

david-vanderson:
Ah okay I didn't get that before.  But won't that mean the update() function (in www/template.html) will be spinning as fast as possible even if the app doesn't need to update anything?

david-vanderson:
Makes sense.  Does having a spinning loop like that mean that not active tabs with mach apps will spin?  Or do browsers pause the requestAnimationFrame stuff for background tabs?  (I'm imagining having a bunch of mach apps in tabs doing things like syncing todo lists)

iddev5:
I think it will still spin, rAF will not pause since there's still some code to process but since it will not get any event, it won't be updated.

slimsag:
Seems browser dependent. 

* Chrome: requestAnimationFrame is paused when the tab is inactive.
* Firefox: requestAnimationFrame runs exponentially slower when the tab is inactive, with each frame taking 1s, 2s, 4s, 8s and so on.
* Edge: requestAnimationFrame is always paused in inactive tabs.
* Safari: requestAnimationFrame is paused as well.

According to https://stackoverflow.com/a/16033979

david-vanderson:
Thanks for the info!  I'll rework it.

www/template.html
```diff
@@ -25,22 +25,17 @@
...
-            instance.exports.wasmUpdate();
-            if (mach.wait_event_timeout > 0) {{
-              mach.wait_event_timer = setTimeout(() => {{
-                window.requestAnimationFrame(update);
-                }},
-                mach.wait_event_timeout * 1000);
-            }}
-            else {{
-              window.requestAnimationFrame(update);
+            if (mach.machHasEvent() ||
```
iddev5, line 31: This is perfect, but with one problem is this will always wait, as in we current can't just poll without waiting. This needs the same type of 3 way comparision as native:
```zig
if (mach.wait && (mach.machHasEvent() || (...)) {

} else {
    rAF(); // Current case, default, poll but don't wait
}
```
The ``mach.wait`` field will be set by Engine.setWaitTimeout();

iddev5: No wait @david-vanderson your'e right. Its in OR.

iddev5 (approved):
Everything works well. This is amazing!

david-vanderson:
Thanks for helping me through it!

slimsag (approved):

PR #332, opened 2022-06-05
mach: Listen to resize event for canvas and thus store window and framebuffer size

iddev5:
Implements #320 for Wasm.


iddev5:
The approach taken in this PR may look a bit complicated because:
1) JS doesnt provide any way to get resize events for DOM elements (The HTML "resize" event is for window object only)
2) A new internal function pollChanges is used (and changes are differentiated from events) because calling pollEvent is optional for user and so code which needs to be updated every frame needs to be placed somewhere else.
3) It is designed in such a way that in future under pollChanges, we can have callbacks for specific changes instead of trying to do it every frame regardless of size change (for example swap chain reconfiguration)

slimsag (approved):
This looks excellent, nice call on using MutationObserver. Thanks for your attention to detail here!

There are some conflicts before I can merge, can you rebase?

iddev5:
Rebased.

PR #334, opened 2022-06-07
ECS fixes

dweiller:
Two small fixes for the ECS

- Fix the usage of `copy()`, which had the `src_row` and `dst_row` arguments supplied in the wrong order
- Fix a bug cause by pointer invalidation in the `setComponent()` and `removeComponent()` functions, caused by `getOrPut()` on the hash map sometimes invalidating the archetype retrieved by `archetypeByID()`.

I haven't made a minimal reproducible example for the second bug, it doesn't show up under all circumstances, in my use it showed up when 'enough' stuff was added to the ECS, which must have made `getOrPutContext()` more likely to invalidate the pointer.

Perhaps there is a better way to rewrite the usage of `archetype` in the functions, this is just the minimal change that stops things crashing.

### DIFF
--- a/ecs/src/entities.zig
+++ b/ecs/src/entities.zig
@@ -371,6 +371,10 @@ pub const Entities = struct {
         // new component was added), or the same archetype storage table (if just updating the
         // value of a component.)
         var archetype_entry = try entities.archetypes.getOrPut(entities.allocator, new_hash);
+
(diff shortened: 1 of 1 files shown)


slimsag (approved):
Thanks! I have more major changes to this ECS library coming soon. Interesting to hear you are using it for something! In general (and as you have found) I would say it's not ready for that yet :)

PR #336, opened 2022-06-08
freetype: Avoid dereferencing null bitmap buffer.

jamii:
For some characters (eg \r) the returned glyph will have .{.rows = 0, .width=0, .pitch = 0, .buffer = null}. In zig null[0..0] causes a panic rather than returning an empty slice.


alichraghi:
thanks @jamii 
i think it would be better to return an optional (`?[]const u8`) instead of empty buffer
this also requires a minor change in *single-glyph* example

alichraghi:
@jamii this is a binding so it must be user choice to unwrap optional values (no self though code as much as possible)

alichraghi (approved):

slimsag (approved):
Thanks a ton!

PR #340, opened 2022-06-11
mach: Implement fullscreen application support

iddev5:
Adds a new field to mach.Options called ``fullscreen: bool = false``. This PR implements fullscreen support for native and wasm platforms. 

For native, glfw doesn't supports getting the currently active monitor (i.e monitor having input focus), so the fullscreen window opens up on primary window by design.

For wasm, fullscreen support is primarily implemented with styles because html fullscreen support is broken in many ways. With fullscreen support, the border around canvas is also slightly broken when you switch between windows and fullscreen modes but thats a problem only for as long as we dont have rendering support for wasm.

---


slimsag (approved):
Great!

PR #342, opened 2022-06-11
harfbuzz: more coverage

alichraghi:

### DIFF
--- a/freetype/src/harfbuzz/blob.zig
+++ b/freetype/src/harfbuzz/blob.zig
@@ -35,7 +35,7 @@ pub const Blob = struct {
         };
     }
 
-    pub fn getEmpty() Blob {
+    pub fn initEmpty() Blob {
         return .{ .handle = c.hb_blob_get_empty().? };
     }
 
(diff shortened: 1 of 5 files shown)


freetype/src/harfbuzz/blob.zig
```diff
@@ -13,13 +13,13 @@ pub const Blob = struct {
 
     pub fn init(data: []u8, mode: MemoryMode) ?Blob {
         return Blob{
-            .handle = c.hb_blob_create_or_fail(data.ptr, @intCast(c_uint, data.len), @enumToInt(mode), null, null) orelse return null,
+            .handle = c.hb_blob_create_or_fail(&data[0], @intCast(c_uint, data.len), @enumToInt(mode), null, null) orelse return null,
```
slimsag, line 16: FYI this means an empty `data` slice cannot be passed in (panic on accessing index 0). I assume that's what you want / this is intentional? Otherwise `.ptr` would do what you want when `data.len == 0`.

slimsag (approved):

freetype/src/harfbuzz/blob.zig
alichraghi: yes i just though it's better to panic there

PR #352, opened 2022-06-14
mach: mouse cursor support

david-vanderson:

### DIFF
--- a/src/platform/native.zig
+++ b/src/platform/native.zig
@@ -20,7 +20,7 @@ pub const Platform = struct {
     last_position: glfw.Window.Pos,
     wait_event_timeout: f64 = 0.0,
 
-    cursors: [@typeInfo(enums.MouseCursor).Enum.fields.len]glfw.Cursor = undefined,
+    cursors: [@typeInfo(enums.MouseCursor).Enum.fields.len]?glfw.Cursor = undefined,
 
     native_instance: gpu.NativeInstance,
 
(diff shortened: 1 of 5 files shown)


david-vanderson:
I'm working on the web backend support for this too.  Unsure how custom cursors should be handled.  I'm not trying to make custom cursors work yet but don't want to prevent it with a bad API.  Has anyone thought about how this should look?

src/platform/native.zig
```diff
@@ -173,6 +175,18 @@ pub const Platform = struct {
...
+        mouse_cursors[@enumToInt(enums.MouseCursor.arrow)] = try glfw.Cursor.createStandard(.arrow);
+        mouse_cursors[@enumToInt(enums.MouseCursor.ibeam)] = try glfw.Cursor.createStandard(.ibeam);
+        mouse_cursors[@enumToInt(enums.MouseCursor.crosshair)] = try glfw.Cursor.createStandard(.crosshair);
+        mouse_cursors[@enumToInt(enums.MouseCursor.pointing_hand)] = try glfw.Cursor.createStandard(.pointing_hand);
+        mouse_cursors[@enumToInt(enums.MouseCursor.resize_ew)] = try glfw.Cursor.createStandard(.resize_ew);
+        mouse_cursors[@enumToInt(enums.MouseCursor.resize_ns)] = try glfw.Cursor.createStandard(.resize_ns);
+        mouse_cursors[@enumToInt(enums.MouseCursor.resize_nwse)] = try glfw.Cursor.createStandard(.resize_nwse);
+        mouse_cursors[@enumToInt(enums.MouseCursor.resize_nesw)] = try glfw.Cursor.createStandard(.resize_nesw);
+        mouse_cursors[@enumToInt(enums.MouseCursor.resize_all)] = try glfw.Cursor.createStandard(.resize_all);
+        mouse_cursors[@enumToInt(enums.MouseCursor.not_allowed)] = try glfw.Cursor.createStandard(.not_allowed);
```
slimsag, line 188: Not all of these cursors are available on all platforms with GLFW, and so all these `try` statements will actually make `init` just outright fail on some systems (primarily Linux and macOS)

What we'll need to do is handle cursor requests in a similar way to GLFW, calling `createStandard` only when needed and storing the result, and properly returning an error in the event the cursor is not available.

slimsag:
@david-vanderson thanks so much for also working on the web backend, I really appreciate you keeping the feature parity going as much as possible.

https://electrictoolbox.com/change-cursor-javascript/ gives some straightforward examples of how we can set the cursor, and it should be very easy to map these to the GLFW cursor values too. I think the API you have here (aside from lack of error handling) should work perfectly fine in the wasm context. Just need to call a JS shim and set `document.body.style.cursor`

src/platform/native.zig
slimsag: Alternatively, we could say for now that if setting the cursor fails we make it a no-op (ignore the error or log it), and in the future produce custom cursor images with GLFW for any OS that does not have native cursors for these.

I kind of like that idea a little better, because then it reduces the chances someone does **exactly this**: think they're available everywhere, and write an application that just panics/crashes on linux/macos by using `try`.

Thoughts?

david-vanderson:
Thanks!  And thanks for the pointers - looks straightforward.

david-vanderson:
I'm sure the web version could be better but at least it works.

slimsag:
Rather than invoking `glfw.Cursor.createStandard` up-front for all cursors, could we invoke it only once a custom cursor is requested if we have not done so before? e.g. if `mouse_cursors[N]` is null, try the `createStandard` call again?

The benefit of this is that it reduces memory a little, and reduces time spent on startup a little.

slimsag:
> Oh totally forgot - Do we need to destroy the glfw cursors created? I noticed there's no platform.deinit(). Is that handled in some other way?

Good question! Yes, we do need to destroy them. We can add a `defer engine.internal.deinit()` call here: https://github.com/hexops/mach/blob/7df25afc6251cd6180708f2dbc97455b903dc920/src/platform/native.zig#L533-L537

src/platform/native.zig
```diff
@@ -318,6 +337,19 @@ pub const Platform = struct {
         return platform.last_window_size;
     }
 
+    pub fn setMouseCursor(platform: *Platform, cursor: enums.MouseCursor) !void {
+        const glfw_cursor = platform.cursors[@enumToInt(cursor)];
+        if (glfw_cursor) |cur| {
+          try platform.window.setCursor(cur);
+        }
+        else {
```
slimsag, line 345: Ah also make sure you run `zig fmt`

david-vanderson:
Hope that helps.  Sorry I missed zig fmt.  I have a list of bools recording if we already tried a particular cursor enum, so we aren't constantly trying and failing.  I think it makes sense but let me know if you'd rather it work some other way?

src/platform/native.zig
```diff
@@ -318,6 +330,39 @@ pub const Platform = struct {
...
+                .resize_nesw => glfw.Cursor.createStandard(.resize_nesw) catch null,
+                .resize_all => glfw.Cursor.createStandard(.resize_all) catch null,
+                .not_allowed => glfw.Cursor.createStandard(.not_allowed) catch null,
+            };
+        }
+
+        if (platform.cursors[enum_int]) |cur| {
+            try platform.window.setCursor(cur);
+        } else {
+            // In the future we shouldn't hit this because we'll provide backup
```
slimsag, line 359:
```suggestion
            // TODO: In the future we shouldn't hit this because we'll provide backup
```

slimsag (approved):
Excellent, this looks great - thanks so much!

PR #354, opened 2022-06-15
glfw: submodule path fix for subrepo

Vulfox:

I've been using the mach-glfw sub(cloned)-repo and the recent submodule path change from https://github.com/hexops/mach/pull/346 doesn't match up with what mach-glfw expects as a standalone repo. This should be reproducible with the example vulkan triangle repo: https://github.com/hexops/mach-glfw-vulkan-example.


slimsag (approved):
Thank you!

PR #359, opened 2022-06-19
ecs: allow components to have type void

dweiller:
Allow size 0 (void) components. These can be used as flags allowing to iterate over entities with certain flags, but without the need to branch on the value of a bool/bitset, ~but there is an extra branch in the get/set functions~ the new branch in get/set should be comptime evaluated and not impose any overhead at runtime. This also doesn't add any per-entity storage overhead, but does add the overhead of new archetypes for each combination of void-components used.


slimsag (approved):
OK nice, I like this, it's exactly how I was envisioning this would work. Great work, thanks!

Issue #360, opened 2022-06-20
mach: allow setting magical declarations used by the zig standard library


dweiller:
The discussion in #309 says that Mach won't use magical public declarations to set things such as options, however I think it makes sense to use them for (any applicable) magical public declarations that `std` looks for. This shouldn't impose any extra cognitive overhead compared to what `std` already does and not re-exporting them may make somethings very hard; for example at the moment if you want to `std.log` with a custom logging function or logging level you have to modify `src/platform/{native,wasm}.zig`.

Grepping Zig's standard library for 'hasDecl\(root' suggests that the magical declarations that might makes sense to export are:

  - `log`
  - `log_level`
  - `scope_levels`
  - `allow_todo_in_release`
  - `panic`
  - `enable_segfault_handler`
  - `event_loop`
  - `io_mode`

There are roughly in order of how clear I think it is that they be re-exported by Mach `"root"` package. I would default to saying all of these should be re-exported unless Mach wants to impose its own choice (e.g. maybe Mach just has its own event loop and doesn't let you use your own with `std`).

The other magic decls that `std` looks for are either `os` or ones used by startup code (e.g. `main`, `_start`, `WinMainCTRStartup` etc), which presumably don't need to be considered.

I'm not totally sure how to handle `log` for wasm.

iddev5:
Mach provides a log for wasm. But it cannot be that easily replaced. Providing an alternative reimplementation for the user would be complicated. Same goes for panic for wasm.

I don't see why this would be useful. Mach being a high level application framework should always provide the most sensible defaults out of the box.

slimsag:
I think it's very important we consider each of these in isolation, not all together. Merely re-exporting all of these is probably not a good idea.

It's important that we consider the use case of each, how they apply to different platforms, and consequences of supporting them.

## For logging in specific

I don't know enough about Zig's logging library, I haven't read it's source yet, but the full impact of this change is not clear to me. Presumably `std.log.info`, `std.log.err`, etc. work - but it sounds like they're not sufficient from your POV?

The other thing here is, beyond WASM support (which also needs to be considered), is this:

> a custom logging function

How exactly we support this needs a lot of thought, especially defaults. Once we have an editor, for example, I can imagine it being quite useful to have logs show up in the editor. In contrast, for most release-mode applications it seems likely that you'd want logs to go to a file - because someone opening your `.exe` isn't going to be storing those for you in the event something goes wrong. Crash reporting is another thing in this area.

[...]

PR #361, opened 2022-06-20
Use `log_level` and `scope_levels` pub decls from App

dweiller:
This PR makes `src/platform/native.zig` re-export logging declarations from the `"app"` package. As mentioned in #360, these are the magical declarations `std` uses that I think should clearly be re-exported.

I'm not sure at the moment to handle a custom `log` function for wasm as I haven't used it, so this PR only deals with the native target.

### DIFF
--- a/src/platform/native.zig
+++ b/src/platform/native.zig
@@ -8,6 +8,9 @@ const enums = @import("../enums.zig");
 const util = @import("util.zig");
 const c = @import("c.zig").c;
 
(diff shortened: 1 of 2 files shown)


dweiller:
Given the discussion in #360 I would be perfectly happy to remove the re-exporting of `log`, I'm mostly interested in `log_levels` and `scope_levels` at the moment.

dweiller:
I've updated this PR to re-export only `log_level` and `scope_levels`, on both wasm and native targets.

slimsag (approved):

Issue #357, opened 2022-06-16
Proposal: Resource management system


iddev5:
## Description
General purpose resource management engine which is meant to be cross platform (works on desktop, wasm, android etc.)

The general idea is that in order to access resources, we will use URIs instead of path. All resources will be accessed from the provided API and std.fs.\* stuffs should not be used (but the library may use it internally).

The URI will internally evaluate to a path (for desktops) or to an URL (for web). On desktops it will use std.fs APIs and on web it will stream resources which are served on your web server.

## Mechanism
Let's say the projects needs a.png, b.png and c.mp3. Then the resources have to be arranged in your directory as such:
- data/
	- textures/
		- a.png
		- b.png
	- audio/
		- c.mp3

This arrangement can be done automatically by some code, which takes a config file (json/yaml/toml) as input and sort the files by their names. The file can look like as such (using json for example):
```json
{
	"collections": [
		{
			"name": "textures",
			"resources": [
				"a.png",
				"b.png"
			]
		},
		{
			"name": "audio",
			"resources": [
				"c.mp3"		
			]
		}
	]
}
```

The resources arrangement/installation can be part of a build step for seamless building. This step may just install or optionally generate an archive file from all the resources.

The resources will be referred inside the application as such: ``textures/a.png``, ``textures/b.png``, ``audio/c.mp3``. But why not ``data/textures/a.png``? (Answered below in API)

Remember that the names ``textures``, ``audio`` etc above are totally cosmetic and has no significance. It can be arbitrarily named anything. [1]

## API
The library will provide the following API:
```zig
// The main high level API
pub const ResourceManager = struct {
    pub fn init(resource_types: []const ResourceType) ResourceManager;

    pub const ResourceType = struct {
        name: []const u8,
        ext: []const []const u8, // [2]
        is_binary: bool, // May not be needed
        load: *const fn (mem: []const u8) !*anyopaque,
        unload: *const fn (resource: *anyopaque) void,
    };

    // Loads ``name`` collection to the resource manager.
    //
    // ``name`` is the address of the particular collection. This is relevant in context of
    // when we have multiple archives. This name will be used when accessing the resources:
    // ``{name}://textures/abc.png``.
    //
    // The ``root`` param will dictate from where will the resources be loaded, by default
    // it is set to ``{appdir}/data/``. This provides us with the convenience of being able
    // support different mods/themes/resource packs for the end user. It can also point to
    // a file (archive), like ``{appdir}/data.mres``
[...]

PR #362, opened 2022-06-21
mach: Begin resource management system

iddev5:
Foundation for #357 

Implements a basic resource management system with proper processing of resources. So far supporting only native backends. It doesn't contain grouping, wasm support, multithreading etc. and they will be part of a future PR.

Currently no example has been ported to this API, but it did worked with basic data lifetime and context testing.

---


slimsag (approved):
Looks like a great starting point, thank you!

Issue #180, opened 2022-03-09
build: self-hosted compiler progress


slimsag:
This is the tracking issue for getting all Mach examples/libraries working with the new self-hosted Zig compiler, which will be enabled by default in Zig v0.11 and is already enabled by default in all nightly Zig builds.

## Status

* aarch64-macos: tested 2022-09-09
* x86_64-linux: tested 2022-09-08
* x86_64-windows: tested 2022-09-09

| aarch64-macos | linux | windows | test                                                                        |
| ------------- | ----- | ------- | --------------------------------------------------------------------------- |
| ✅             | ✅     | ✅       | `cd libs/ecs && zig build test -fno-stage1`                                 |
| ✅             | ✅     | ✅       | `cd libs/ecs && zig build test -fstage1`                                    |
| ✅             | ✅     | ✅       | `cd libs/gpu && zig build test`                                             |
| ✅             | ✅     | ✅       | `cd libs/gpu && zig build run-example`                                      |
| ✅             | ✅     | ✅       | `cd libs/glfw && zig build test`                                            |
| ✅             | ✅     | ✅       | `cd libs/sysaudio && zig build test`                                        |
| ✅             | ✅     | ✅       | `cd libs/sysaudio && zig build run-example-soundio-sine-wave`               |
| ✅             | ✅     | ✅       | `cd libs/freetype && zig build test`                                        |
| ✅             | ✅     | ✅       | `cd libs/freetype && zig build run-example-single-glyph -- 'a'`             |
| ✅             | ✅     | ✅       | `cd libs/freetype && zig build run-example-glyph-to-svg`                    |
| ✅             | ✅     | ✅       | `cd libs/gpu-dawn && zig build`                                             |
| ✅             | ✅     | ✅       | `zig build-exe --main-pkg-path . ./tools/html-generator.zig`                |
| 🍎             | 🍎     | 🍎       | `zig build run-example-triangle -fstage1 -Dtarget=wasm32-freestanding-musl` |
| ✅             | ✅     | ✅       | `zig build run-shaderexp`                                                   |
| ✅             | ✅     | ✅       | `zig build run-example-advanced-gen-texture-light`                          |
| ✅             | ✅     | ✅       | `zig build run-example-boids`                                               |
| ⚠️             | ⚠️     | ⚠️       | `zig build run-example-ecs-app`                                             |
| ✅             | ✅     | ✅       | `zig build run-example-fractal-cube`                                        |
[...]

PR #365, opened 2022-06-23
glfw: support compiling glfw with stage2

PiergiorgioZagaria:


PiergiorgioZagaria:
One strange thing I noticed is that -fno-stage1 sees 137 tests, while normally there are 138

PiergiorgioZagaria:
I wrote about it in #180 basically stage2 has problems recognising when to call something at comptime, and there are breaking changes in how translate-c translates string pointers

slimsag (approved):
Very nice, thanks for doing this @PiergiorgioZagaria !

Issue #183, opened 2022-03-17
gpu: verify lifetimes of descriptor memory, investigate alternatives to copying strategy


slimsag:
In a few places we have to copy descriptors & structs before passing into the native implementation because our interface is not always ABI compatible with `webgpu.h`:

https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/hexops/mach%24%40webgpu+file:gpu+page_allocator&patternType=literal

We're currently using a pretty gnarly approach: we use stack space if the data we have to copy is small enough, and heap-allocate otherwise. Obviously, the memory is always gone (stack or not) once the function returns. It's not exactly clear from `webgpu.h` what the lifetime of certain memory is, as there are no docs. I believe descriptors must live only until the function which takes the descriptor returns, but this needs to be confirmed.

- [ ] Confirm all memory lifetimes are valid
- [ ] Identify if there is a better / more optimal strategy for achieving ABI compatibility so no copying is needed.
- [ ] "The `gpu.Queue` was going out of scope before the callback got called, so the wrapper callback couldn't call the `gpu.Queue.WorkDoneCallback` it held" from [this matrix chat](https://matrix.to/#/!bIJeTLtOHucXQpqIcH:matrix.org/$AZSS1Q2njUjiH4lRbH7KvsaEyqqUITJT0SHQPInHWjc?via=matrix.org&via=mozilla.org&via=pythons.space)

slimsag:
https://github.com/hexops/mach/pull/403

PR #366, opened 2022-06-23
Store pointer to WorkDoneCallback in gpu.Queue

dweiller:
This change removes a footgun where the queue used to submit with a callback must be valid when the callback is run.  This change makes it clear (hopefully) to the user that the queue does not need to be valid, only the `WorkDoneCallback` does, when the callback gets run.

Helps hexops/mach#183


slimsag (approved):
Thanks for this, much appreciated!

PR #374, opened 2022-06-28
glfw: build for wayland *and* X11 by default

PiergiorgioZagaria:
Helps hexops/mach#374 and hexops/mach#376


PiergiorgioZagaria:
To get this error I set `x11 = false` in the build options, otherwise glfw would use X11 instead of wayland.
Also, all tests from 10 to 138 print `mesa: for the --amdgpu-atomic-optimizations option: may only occur zero or one times!`.

The errors are:
- `Test [46/138] Monitor.test "set_getGammaRamp"` the documentation seems to be wrong, since on wayland the returned error is `FeatureUnavailable` and not `PlatformError`
- `Test [49/138] key.test "getName"`, the c function segfaults on wayland
```
Segmentation fault at address 0x80
???:?:?: 0x7f683544b9c4 in ??? (???)
upstream/glfw/src/input.c:713:12: 0x30a0a7 in glfwGetKeyName (/home/Zargio/Documents/Github/mach/glfw/src/sources_all.c)
    return _glfw.platform.getScancodeName(scancode);
           ^
/home/Zargio/Documents/Github/mach/glfw/src/key.zig:220:43: 0x248115 in key.test "getName" (test)
        const name_opt = cc.glfwGetKeyName(@enumToInt(self), @intCast(c_int, scancode));
```
- `Test [81/138] Window.test "setAspectRatio"` returns error `FeatureUnimplemented`
- `Test [89/138] Window.test "show"` like the first error, `FeatureUnavailable` is returned instead of `PlatformError
- `Test [92/138] Window.test "requestAttention"` returns error `FeatureUnimplemented`

slimsag:
> the c function segfaults on wayland

I suspect this is more undefined behavior being caught by Zig/UBSan, I can dig into it later. Presumably it works if you do a release-fast build

slimsag:
Let's just comment it out for now and add:

```
// TODO: https://github.com/hexops/mach/issues/375
```

slimsag (approved):
Looks good, thanks a ton again for doing this!

slimsag:
CI shows this is failing to cross-compile from macOS/Windows:

image

It looks like linking to `wayland-client` is no longer necessary, though, because GLFW loads it at runtime now. I am pretty sure we can just remove that line entirely from build.zig?

slimsag (approved):
LGTM, thanks! I did a `git rebase -i origin/main` and force pushed to clean up the history a bit, hope that's okay.

Awesome work!

PR #377, opened 2022-06-28
mach: flesh out mach.Event

jamii:
This adds some additional events. It also does some busywork tracking cursor position and key modifiers and attaches copies of that state to events.

I don't see any tests for mach platform but `zig build compile-all` works and keypresses still work in `zig build run-example-advanced-gen-texture-light`

### DIFF
--- a/src/platform/mach.js
+++ b/src/platform/mach.js
@@ -196,28 +196,40 @@ const mach = {
     canvas.addEventListener("contextmenu", (ev) => ev.preventDefault());
 
     canvas.addEventListener("keydown", (ev) => {
-      mach.events.push(...[1, convertKeyCode(ev.code)]);
+        if (ev.repeat) {
+            mach.events.push(...[2, convertKeyCode(ev.code)]);
+        } else {
+            mach.events.push(...[1, convertKeyCode(ev.code)]);
+        }
     });
 
     canvas.addEventListener("keyup", (ev) => {
-      mach.events.push(...[2, convertKeyCode(ev.code)]);
+      mach.events.push(...[3, convertKeyCode(ev.code)]);
     });
 
     canvas.addEventListener("mousemove", (ev) => {
-      mach.events.push(...[3, ev.clientX, ev.clientY]);
+      mach.events.push(...[4, ev.clientX, ev.clientY]);
 	});
 
     canvas.addEventListener("mousedown", (ev) => {
-      mach.events.push(...[4, ev.button]);
-	});
-
-    canvas.addEventListener("mouseup", (ev) => {
       mach.events.push(...[5, ev.button]);
 	});
 
-    canvas.addEventListener("wheel", (ev) => {
-      mach.events.push(...[6, ev.deltaX, ev.deltaY]);
+    canvas.addEventListener("mouseup", (ev) => {
+      mach.events.push(...[6, ev.button]);
 	});
+
+    canvas.addEventListener("wheel", (ev) => {
+      mach.events.push(...[7, ev.deltaX, ev.deltaY]);
+	});
+
+    canvas.addEventListener("focus", (ev) => {
+        mach.events.push(...[8]);
+    });
+
+    canvas.addEventListener("blur", (ev) => {
+        mach.events.push(...[9]);
+    });
 
     canvas.addEventListener("mach-canvas-resize", (ev) => {
       const cv_index = mach.canvases.findIndex((el) => el.canvas === ev.currentTarget);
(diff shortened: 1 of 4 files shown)


src/platform/wasm.zig
```diff
@@ -53,6 +56,20 @@ pub const Platform = struct {
                 .width = js.machCanvasGetFramebufferWidth(id),
                 .height = js.machCanvasGetFramebufferHeight(id),
             },
+
+            // TODO initialize these properly
```
slimsag, line 60: What does this mean? Initialize them to their current state? How hard is that to do?

src/platform/wasm.zig
```diff
@@ -117,38 +134,84 @@ pub const Platform = struct {
...
+                }
+                break :key_press structs.Event{
+                    .key_press = .{
+                        .key = key,
+                        .mods = platform.last_key_mods,
+                    },
+                };
             },
-            2 => structs.Event{
-                .key_release = .{ .key = @intToEnum(enums.Key, js.machEventShift()) },
```
slimsag, line 128: This switch now no longer handles 2, it jumps from 1 to 3. Why is that?

src/structs.zig
```diff
@@ -48,25 +48,49 @@ pub const Options = struct {
...
     },
+    mouse_press: MouseButtonEvent,
+    mouse_release: MouseButtonEvent,
     mouse_scroll: struct {
         xoffset: f32,
         yoffset: f32,
     },
+    focus_gained,
+    focus_lost,
+    window_closed,
```
slimsag, line 68: Should `window_closed` be renamed to just `closed`?

src/platform/mach.js
```diff
@@ -200,25 +200,33 @@ const mach = {
     });
 
     canvas.addEventListener("keyup", (ev) => {
-      mach.events.push(...[2, convertKeyCode(ev.code)]);
+      mach.events.push(...[3, convertKeyCode(ev.code)]);
```
slimsag, line 203: Would love to understand why these are changing. Probably I've just missed something in this PR.

slimsag:
Looks great overall, thanks a ton for doing this! I left some questions inline.

src/platform/wasm.zig
jamii: I can add code to the js side to do this and plumb it through, but I'm reluctant to make non-trivial changes to the js because none of the wasm examples work for me so I have no way to test whether I'm writing gibberish or not.

src/platform/wasm.zig
jamii: I didn't add js code to handle key_repeat (same reason as above). This one is a pretty trivial change though so probably ok.

src/platform/mach.js
jamii: Same as above - there is a new key_repeat event to match glfw.

jamii:
I tried using modifier states from key/mouse events as per https://matrix.to/#/!hPlvgnLQhMkTucKPrh:matrix.org/$QKzo9E6XZi5yY9jEYbU8BbCPBs66IjyI9Pr5X5hhzXc?via=matrix.org&via=scattered-thoughts.net, which would avoid the problem of initializing mods.

But this often gives inconsistent results eg I found that pressing right alt did not trigger ev.altKey or ev.getModifierState("AltGraph"). Additionally, the names of keys varies between browsers - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState#browser_compatibility.

SImilarly, pressing the windows key on my keyboard produces a left_super event but does not set ev.getModifierState("OS") or ev.getModifierState("Meta").

So I think it's probably better to stick with the current code, so that at least the mods are consistent with the stream of events that the dev sees.

jamii:
The CI failures are timeouts fetching dependencies.

src/structs.zig
slimsag: let's drop the `window` prefix, and just go with `closed`

src/platform/wasm.zig
slimsag: Got it, let's do that in a follow-up PR.

slimsag:
Yeah, ignore the CI failures, I'll look into those later.

@iddev5 do you mind reviewing when you have a chance?

iddev5 (approved):

iddev5:
> I tried using modifier states from key/mouse events as per

Thanks for looking into it. The current approach would work for now.

slimsag (approved):

PR #379, opened 2022-06-29
audio: make `buildSoundIo` private and point package to `./main.zig`

alichraghi:

### DIFF
--- a/audio/build.zig
+++ b/audio/build.zig
@@ -4,7 +4,7 @@ const Builder = std.build.Builder;
 const soundio_path = thisDir() ++ "/upstream/soundio";
 
 pub const pkg = std.build.Pkg{
-    .name = "mach-audio",
+    .name = "audio",
     .source = .{ .path = thisDir() ++ "/main.zig" },
 };
 
--- a/audio/build.zig
+++ b/audio/build.zig
@@ -4,16 +4,8 @@ const Builder = std.build.Builder;
 const soundio_path = thisDir() ++ "/upstream/soundio";
 
 pub const pkg = std.build.Pkg{
-    .name = "soundio",
-    .source = .{ .path = thisDir() ++ "/soundio/main.zig" },
-};
-
-pub const SoundIoOptions = struct {
-    jack: bool = false,
-    pulseaudio: bool = false,
-    alsa: bool = false,
-    coreaudio: bool = false,
-    wasapi: bool = false,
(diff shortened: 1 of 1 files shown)


audio/build.zig
```diff
@@ -4,16 +4,8 @@ const Builder = std.build.Builder;
...
-    .source = .{ .path = thisDir() ++ "/soundio/main.zig" },
-};
-
-pub const SoundIoOptions = struct {
-    jack: bool = false,
-    pulseaudio: bool = false,
-    alsa: bool = false,
-    coreaudio: bool = false,
-    wasapi: bool = false,
+    .name = "mach-audio",
```
slimsag, line 7:
```suggestion
    .name = "audio",
```

I don't know if `mach-audio` is better, but `glfw` and `freetype` are not `mach-glfw` or `mach-freetype` so it seems clear to me this should just be `audio` for consistency sake. Open to changing them all though!

audio/build.zig
```diff
@@ -22,19 +14,19 @@ pub fn build(b: *Builder) void {
...
-    link(b, soundio_tests, .{ .alsa = true });
+    link(b, soundio_tests);
 
     const test_step = b.step("test", "Run library tests");
     test_step.dependOn(&soundio_tests.step);
 }
 
-pub fn link(b: *Builder, step: *std.build.LibExeObjStep, options: SoundIoOptions) void {
-    const soundio_lib = buildSoundIo(b, step, options);
+pub fn link(b: *Builder, step: *std.build.LibExeObjStep) void {
```
slimsag, line 23: I do think we should still take an options struct in this function parameter, just so we can add any options necessary in the future without breaking user's current use of `audio.link(...)`. It should just be an empty `Options` struct for now, though.

audio/build.zig
```diff
@@ -4,16 +4,8 @@ const Builder = std.build.Builder;
 const soundio_path = thisDir() ++ "/upstream/soundio";
 
 pub const pkg = std.build.Pkg{
-    .name = "soundio",
-    .source = .{ .path = thisDir() ++ "/soundio/main.zig" },
-};
-
-pub const SoundIoOptions = struct {
```
slimsag, line 11: I would be OK with keeping this struct around, just making it private - if you think that's better. I think this way is fine though.

slimsag (approved):

PR #382, opened 2022-06-30
audio: make soundio binding good enough to run sine-wave example

alichraghi:


slimsag (approved):
Nice work! This is exciting :D I pushed a few commits to get this building on macOS too just now. Which OS do you use/test on? I'll test on the third.

alichraghi:
thanks! atm im running Linux Mint.

PR #388, opened 2022-07-06
freetype: support compiling with stage2 (-fno-stage1)

PiergiorgioZagaria:

Two things I wanted to notice are, utils_pkg (see https://github.com/hexops/mach/issues/180#issuecomment-1175993008) and test_option, `@embedFile` can't import from `../` so we need to use `addOption` in the build file. Also we can't `@embedFile` in build.zig because when running build.zig for the first time upstream/ hasn't been downloaded yet.

### DIFF
--- a/freetype/build.zig
+++ b/freetype/build.zig
@@ -67,11 +67,20 @@ pub fn build(b: *std.build.Builder) !void {
     main_tests.setBuildMode(mode);
     main_tests.setTarget(target);
     main_tests.addPackage(c_pkg);
+
+    // Remove once the stage2 compiler fixes pkg std not found
+    main_tests.addPackage(utils_pkg);
+
     main_tests.addPackage(pkg);
     link(b, main_tests, .{ .freetype = .{
         .ft_config_path = "./test/ft",
         .brotli = true,
     } });
+    var font_option = b.addOptions();
+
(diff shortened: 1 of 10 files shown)


alichraghi (approved):

freetype/build.zig
```diff
@@ -67,11 +67,20 @@ pub fn build(b: *std.build.Builder) !void {
...
+    main_tests.addPackage(utils_pkg);
+
     main_tests.addPackage(pkg);
     link(b, main_tests, .{ .freetype = .{
         .ft_config_path = "./test/ft",
         .brotli = true,
     } });
+    var font_option = b.addOptions();
+
+    font_option.addOption([]const u8, "font", try std.fs.cwd().readFileAlloc(b.allocator, "upstream/assets/FiraSans-Regular.ttf", std.math.maxInt(u32)));
```
iddev5, line 81: You can just set ``exe.main_pkg_path`` which will solve the @embedFile issue

iddev5 (approved):

freetype/build.zig
PiergiorgioZagaria: So I add `main_tests.main_pkg_path = "upstream/assets/";` in build.zig and use `@embedFile("FiraSans-Regular.ttf")`?
It doesn't work so I don't think that's what you meant

iddev5: I meant setting ``main_pkg_path`` to thisDir() and then @embedFile("../upstream...")

PiergiorgioZagaria: Ah yeah, that was it thanks

slimsag (approved):

PR #390, opened 2022-07-06
js-runtime: add/upstream new js-runtime library inspired by Go's syscall/js package

iddev5:
Upstreams js-runtime from https://github.com/iddev5/zig-js-runtime. This module should be considered experimental and its API is not final. It is inspired from Go's syscall/js module. It is not a reimplementation, but rather an independent implementation.

There is also a workaround for a limitation right now. The html-generator has no way to dynamically add JS sources (and that can't be done without using a preprocessor library) so we hardcode js-runtime in it.

In the future, I think the correct behavior would be to move tools/ inside js-runtime along with a wasm application building SDK and get rid of any direct JS access we have today (which is just src/platform/wasm.zig and src/platform/mach.js).

---

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -5,6 +5,7 @@ const gpu_dawn = @import("gpu-dawn/build.zig");
 pub const glfw = @import("glfw/build.zig");
 pub const ecs = @import("ecs/build.zig");
 const freetype = @import("freetype/build.zig");
+const js_runtime = @import("js-runtime/build.zig");
 const Pkg = std.build.Pkg;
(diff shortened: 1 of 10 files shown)


slimsag (approved):
🔥 I'm so excited to start playing with this more

> In the future, I think the correct behavior would be to move tools/ inside js-runtime along with a wasm application building SDK and get rid of any direct JS access we have today (which is just src/platform/wasm.zig and src/platform/mach.js).

agree

PR #393, opened 2022-07-07
glfw: respect negative coordinates in window

janeptrv:
While attempting to run the mach examples on WSLg, I ran into an issue using the OpenGL backend:
```
thread 11857 panic: attempt to cast negative value to unsigned integer
/home/jane/mach/glfw/src/Window.zig:581:22 0x4af643 in Platform.init (example-boids)
    return Pos{ .x = @intCast(u32, x), .y = @intCast(u32, y) };
                     ^
```

the simplest solution, to use a signed integer instead of an unsigned one, worked perfectly fine to resolve this issue and the example built and ran without issue afterwards.

---

### DIFF
--- a/glfw/src/Window.zig
+++ b/glfw/src/Window.zig
@@ -551,8 +551,8 @@ pub inline fn setIcon(self: Window, allocator: mem.Allocator, images: ?[]Image) 
 }
 
 pub const Pos = struct {
-    x: u32,
-    y: u32,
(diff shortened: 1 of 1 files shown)


iddev5:
I think this is correct, window positions can be negative. Though I think it can safely just be ``i32``

slimsag (approved):
Thanks for this!

PR #394, opened 2022-07-07
Audio: initial interface

alichraghi:

### DIFF
--- a/audio/src/main.zig
+++ b/audio/src/main.zig
@@ -1,0 +1,140 @@
+const std = @import("std");
+const mem = std.mem;
+const testing = std.testing;
+const builtin = @import("builtin");
+const Backend = switch (builtin.os.tag) {
+    .linux,
+    .windows,
+    .macos,
+    .ios,
+    => @import("soundio.zig"),
+    else => @compileError("unsupported os"),
+};
+pub const Error = Backend.Error;
+pub const Device = Backend.Device;
+pub const DeviceIterator = Backend.DeviceIterator;
+
+pub const Mode = enum {
+    input,
+    output,
+};
+
+pub const Format = enum {
+    U8,
+    S16,
+    S24,
+    S32,
+    F32,
+};
+
+pub const DeviceDescriptor = struct {
+    mode: ?Mode = null,
+    format: ?Format = null,
+    is_raw: ?bool = null,
+    channels: ?u8 = null,
+    sample_rate: ?u32 = null,
+    id: ?[:0]const u8 = null,
+    name: ?[]const u8 = null,
+};
+
+const Audio = @This();
+
+backend: Backend,
+
+pub fn init() Error!Audio {
+    return Audio{
+        .backend = try Backend.init(),
+    };
+}
+
+pub fn deinit(self: Audio) void {
+    self.backend.deinit();
+}
+
+pub fn waitEvents(self: Audio) void {
+    self.backend.waitEvents();
+}
+
+pub fn requestDevice(self: Audio, config: DeviceDescriptor) Error!Device {
+    return self.backend.requestDevice(config);
+}
+
+pub fn inputDeviceIterator(self: Audio) DeviceIterator {
+    return self.backend.inputDeviceIterator();
+}
+
+pub fn outputDeviceIterator(self: Audio) DeviceIterator {
+    return self.backend.outputDeviceIterator();
+}
+
+test "list devices" {
+    const a = try init();
+    defer a.deinit();
+
+    var iter = a.inputDeviceIterator();
+    while (try iter.next()) |_| {}
+}
+
+test "connect to device" {
+    const a = try init();
+    defer a.deinit();
+
+    const d = try a.requestDevice(.{ .mode = .output });
+    defer d.deinit();
+}
+
+test "connect to device from descriptor" {
+    const a = try init();
+    defer a.deinit();
+
+    var iter = a.outputDeviceIterator();
+    var device_desc = (try iter.next()) orelse return error.NoDeviceFound;
+
+    const d = try a.requestDevice(device_desc);
+    defer d.deinit();
+}
+
+test "requestDevice behavior: null is_raw" {
+    const a = try init();
+    defer a.deinit();
+
+    var iter = a.outputDeviceIterator();
+    var device_desc = (try iter.next()) orelse return error.NoDeviceFound;
+
+    const bad_desc = DeviceDescriptor{
+        .is_raw = null,
+        .mode = device_desc.mode,
+        .id = device_desc.id,
+    };
+    try testing.expectError(error.InvalidParameter, a.requestDevice(bad_desc));
+}
+
+test "requestDevice behavior: null mode" {
+    const a = try init();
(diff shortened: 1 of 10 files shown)


audio/src/main.zig
```diff
@@ -0,0 +1,120 @@
...
+    mode: ?Mode = null,
+    format: ?Format = null,
+    is_raw: ?bool = null,
+    channels: ?u8 = null,
+    sample_rate: ?u32 = null,
+    id: ?[]const u8 = null,
+    name: ?[]const u8 = null,
+    callback: ?fn (device: Device, frame_count: u32) void = null,
+
+    internal: ?DeviceDescriptorInternal = null,
```
iddev5, line 41: This is a bad idea. This should be strictly kept inside soundio.zig's Device struct

audio/src/soundio.zig
```diff
@@ -0,0 +1,201 @@
...
+const Mode = @import("main.zig").Mode;
+const DeviceDescriptor = @import("main.zig").DeviceDescriptor;
+
+const is_raw_sign = "MD_IS_RAW";
+const input_mode_sign = "MD_INPUT_";
+const output_mode_sign = "MD_OUTPUT";
+
+const Audio = @This();
+
+const DeviceUnion = union(Mode) {
```
iddev5, line 18: Naming is incorrect. These are not device but actually streams

audio/src/soundio.zig
```diff
@@ -0,0 +1,201 @@
...
+                    self.ctx.allocator,
+                    "{s}[{s}][{s}]",
+                    .{
+                        device_desc.id(),
+                        is_raw_sign,
+                        switch (@intToEnum(Mode, device_desc.handle.aim)) {
+                            .input => input_mode_sign,
+                            .output => output_mode_sign,
+                        },
+                    },
```
iddev5, line 58: Why are we storing so much information with id. DeviceDescriptor already has is_raw and mode fields.

audio/src/soundio.zig
```diff
@@ -0,0 +1,201 @@
...
+                ),
+                .name = device_desc.name(),
+                .internal = device_desc,
```
iddev5, line 61: Lets not store it, we should get the device only in ``requestDevice`` as mentioned above.

iddev5 (changes requested):

iddev5:
Lets not forget to mention that this PR depends on https://github.com/slimsag/libsoundio/pull/2 and also https://github.com/hexops/soundio needs to be updated accordingly.

iddev5:
I think the purpose of Descriptor should be just to be able to store all the necessary information about a device which we can use to initialize it. It should not store the device handle itself. That would be a bad design. We already have ``soundio.get(*)DeviceFromID`` which just needs Descriptor.id and Descriptor.is_raw.

audio/src/main.zig
alichraghi: in fact there's no need to move it into somewhere else
i was also doubt to keep it, just waited for reviews

audio/src/soundio.zig
alichraghi: so if `internal` and `parseID` get removed, which is correct strategy?
- `InvalidParams` error, if `id` exist but `mode` and `is_raw` was **null**
- automatically connect to best device, if `id` exist but `mode` and `is_raw` was **null**

iddev5: I think we had discussed that we may want to allow device creation just from id, and if I understand it correctly id doesn't have enough information to uniquely identify the devices. So that does justifies what you did above. But it has its problems. 

The id held by soundio we actually get it from the internal audio server itself, so if we made id a custom format, it would be difficult to use the id in future non-soundio native backends. 

We can in theory make the id format a standard which all backends must adhere to but it has the drawback that id would have to be allocated. Alternatively we can just not allow creating device solely from id but require the complete descriptor. cc: @slimsag

slimsag: > The id held by soundio we actually get it from the internal audio server itself, so if we made id a custom format, it would be difficult to use the id in future non-soundio native backends.

I'm not worried about this. In my view, ID should be clearly marked as "a string of garbage only this backend understands" - nobody except that backend should try to interpret it, and there could be changes to the backend (like if we ever move away from soundio) that invalidate those IDs. It's safe to store it in a file and get the device from that ID, but only using the same backend - and not if the backend changes in the future.

audio/src/main.zig
```diff
@@ -0,0 +1,120 @@
...
+}
+
+test "list devices" {
+    const a = try init(testing.allocator);
+    defer a.deinit();
+
+    var iter = a.inputDeviceIterator();
+    while (try iter.next()) |d| {
+        if (d.id) |id|
+            testing.allocator.free(id);
```
slimsag, line 81: It's a bummer we need to free each ID, though, I don't like that much

alichraghi: relax. this happens once game load

iddev5: How do you know the user would do it only once? Well you cant say it surely.

alichraghi: it's user fault and beyond of our control. user can save to an array and use it where want

iddev5: User cannot really store it since devices may change as the program is running, which is triggered by ``flushEvents()`` in my understanding

alichraghi: hmm right. but this seems the only way to keep id standalone. im almost ok with making it user job to save `is_raw` and `mode` itself but it seems other backends provide a unique id.

alichraghi:
will be ready once this merged: https://github.com/hexops/soundio/pull/1

iddev5 (approved):
LGTM! Techincally Device and DeviceIterator should not be completely opaque but we can always fix that later along with that id allocation issue.

audio/src/main.zig
```diff
@@ -0,0 +1,103 @@
...
+
+test "connect to device from id" {
+    const a = try init(testing.allocator);
+    defer a.deinit();
+
+    var iter = a.outputDeviceIterator();
+    var device_desc = (try iter.next()) orelse return error.NoDeviceFound;
+    defer testing.allocator.free(device_desc.id.?);
+
+    const d = try a.requestDevice(device_desc);
```
slimsag, line 98: My thinking (let's chat about this in Matrix):

We should create a new descriptor here, passing just `.{.id = device_desc.id, .mode = .output}` to `requestDevice`. Otherwise we're not sure if we are testing it looking up device by ID or by full descriptor info.

Additionally, we should create another test which verifies what happens when the ID doesn't match any known device. My thinking here is that if you pass `.{.id = "does-not-match-any-existing-device-id", .mode = .output}` for a reason like:

* The user literally doesn't have that device anymore (they unplugged it)
* We switched from soundio to a non-soundio backend and the ID format changed
* ...

Then `requestDevice` could see clearly that the device with that ID no longer exists, and instead give you one that you would get if you just passed `requestDevice(.{.mode = .output})` without the ID field.

alichraghi:
i have a lot of improvements in mind but let's do it in next PRs

slimsag (approved):

PR #401, opened 2022-07-11
ecs: fix pointer invalidation in get/setComponent

dweiller:
The reference to the old archetype is invalidated by getOrPut() calls of
std.ArrayHashMap. The implementation of std.ArrayHashMap means that
pointers can be invalidated on getOrPut() calls _even if the key exists
in the map_. This means that the reference to the old archetype needs to
be refreshed unconditionally (i.e. not only if the new archetype didn't
exist previously).

Maybe it's worth suggesting upstream that `std.ArrayHashMap` changes so `getOrPut()` doesn't invalidate pointers if the key exists. The issue is that `getOrPut()` calls `ensureTotalCapacity()` to make sure there is a free element, thus invalidating pointers if the map is full.


slimsag (approved):
Wow, interesting, didn't realize this was the case. Thanks for this @dweiller !

PR #402, opened 2022-07-11
ecs: type-safe queries/iteration

dweiller:
Now that we have typed ecs components, it makes sense to get type safety into the query/iteration api. This is a first past that works, but I'm not thrilled with the api (passing the query to the `next()` function rather than the `query()` function that constructs the `Iterator`. To make the components a comptime parameter to `query()` instead of `next()` requires the information about the query to be stored in the iterator, which would mean `Iterator` has to be generic over the components queried. Alternatively the component query could be runtime information, but that would require allocating (to concatenate namespace and component name), or imposed a maximum length for namespace + component name.

How should the api for type-safe queries and iteration look? In the future when multithreading/scheduling is worked on this will probably need revisiting as knowledge of what kinds of queries/iteration is done can be exploited for scheduling, but in the meantime it would be good to have a type-safe way to iterate.

### DIFF
--- a/ecs/src/entities.zig
+++ b/ecs/src/entities.zig
@@ -372,9 +372,37 @@ pub fn Entities(all_components: anytype) type {
             row_index: u32,
         };
 
+        pub const Query = Query: {
+            const namespaces = std.meta.fields(@TypeOf(all_components));
+            var fields: [namespaces.len]std.builtin.Type.UnionField = undefined;
+            inline for (namespaces) |namespace, i| {
+                const component_enum = std.meta.FieldEnum(namespace.field_type);
+                fields[i] = .{
+                    .name = namespace.name,
+                    .field_type = component_enum,
+                    .alignment = @alignOf(component_enum),
+                };
+            }
+
+            // need type_info variable (rather than embedding in @Type() call)
+            // to work around stage 1 bug
+            const type_info = std.builtin.Type{
+                .Union = .{
+                    .layout = .Auto,
+                    .tag_type = std.meta.FieldEnum(@TypeOf(all_components)),
+                    .fields = &fields,
+                    .decls = &.{},
+                },
+            };
+            break :Query @Type(type_info);
+        };
+
+        fn fullComponentName(comptime q: Query) []const u8 {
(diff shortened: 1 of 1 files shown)


slimsag:
Thanks for this. I've yet to get around to looking at querying in depth yet, realistically I think that querying cannot be looked at in isolation. Here's what I mean:

* Modules
* Modules doing message-passing
* Multi-threaded scheduling of system execution
* Queries

These four items must be considered together, because each one impacts the other. So far I've gotten close to completing the first two, and investigating the third, but not yet all of them together.

I'm happy to merge this PR as-is right now, though, because it's a clear improvement over what we have today. Any objection to that?

dweiller:
I've pushed a second commit that fixes API so that you pass the components you want to query to the `query()` function rather than `next()`, by making the `Iterator` type generic over those components.

The reason I made this PR a draft was essentially because of those other parts of Mach that querying needs to interact with and wasn't sure how much thought others have put into the API. I'm happy to have these changes merged - these changes can be a starting point for developing the API as those other aspects are considered/implemented (I guess it would be very hard to get all of it 'right' in one go anyway).

I'm not certain what you mean with queries and Modules (and message passing) impacting each other (unless all you mean is that the comptime known components must match up).

slimsag:
> these changes can be a starting point for developing the API as those other aspects are considered/implemented (I guess it would be very hard to get all of it 'right' in one go anyway).

Yup, that's exactly what I mean :)

> I'm not certain what you mean with queries and Modules (and message passing) impacting each other (unless all you mean is that the comptime known components must match up).

Some examples:

1. Should systems support async execution / make use of Zig async? Multi-threading impacts this.
2. How should message passing work in a multi-threaded context? And in a single-threaded context? Async impacts this.
3. How should querying and interacting with ECS data work across threads? Points 1 and 2 impact this, because it could open up a "let systems fight it out with mutexes, interact with anything in the entire ECS" approach rather than a strict "define what you'll interact with up front" approach.

All this to say, there's a _lot_ of conflicting concerns to address - and as you mentioned they can't all be addressed at once. But it's important to address all of them before we say the API is stabilized at all.

slimsag (approved):

PR #406, opened 2022-07-13
libmach: initial API bindings for mach core

zack466:
This pull request includes:
1. C API bindings for Core Mach, exported from `src/bindings.zig`
2. Changes to `build.zig` in order to build `libmach`
3. Test programs located in `libmach/` to check if `libmach` works as expected

### DIFF
--- a/libmach/test.lisp
+++ b/libmach/test.lisp
@@ -1,0 +1,64 @@
+;; Tests the behavior of `libmach` using Common Lisp's CFFI
+;; This Lisp script is basically a one-to-one translation of `test.c`
+
+(ql:quickload :cffi)
+
+(defpackage :cffi-user
+  (:use :cl :cffi))
+
+(in-package :cffi-user)
+
+(define-foreign-library libmach
+  (t (:default "./build/libmach")))
+   
+(use-foreign-library libmach)
+
+;; Note: CFFI automatically translates C_style names into lispier kebab-case ones
+
+;; typedef void mach_core_callback(void*);
+(defctype mach-core-callback :pointer)
+
+;; void mach_core_set_init(mach_core_callback);
+(defcfun "mach_core_set_init" :void
+  (callback mach-core-callback))
+
+;; void mach_core_set_update(mach_core_callback);
+(defcfun "mach_core_set_update" :void
+  (callback mach-core-callback))
+
+;; void mach_core_set_deinit(mach_core_callback);
+(defcfun "mach_core_set_deinit" :void
+  (callback mach-core-callback))
+
+;; void mach_run(void);
+(defcfun "mach_run" :void)
+
+;; void core_set_should_close(void*);
+(defcfun "core_set_should_close" :void (core :pointer))
+
+;; float core_delta_time(void*);
+(defcfun "core_delta_time" :float (core :pointer))
+
+(defcallback my-init :void ((core :pointer))
+  (format t "Hello from my-init!~%"))
+
+(defvar *elapsed* 0.0)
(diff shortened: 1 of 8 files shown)


slimsag (approved):
This looks like a great starting point, and is a clear improvement over what we have here today so I'm going to merge it. We can continue to iterate on this. I'll leave feedback inline on how we can improve it next

src/bindings.zig
```diff
@@ -0,0 +1,52 @@
...
+    libmach.core_callbacks.core_update = core_update;
+}
+
+pub export fn mach_run() void {
+    if (libmach.core_callbacks.core_init == null) {
+        std.debug.print("Did not provide a core_init callback\n", .{});
+        return;
+    }
+    if (libmach.core_callbacks.core_update == null) {
+        std.debug.print("Did not provide a core_update callback\n", .{});
```
slimsag, line 36: Perhaps we should panic in these cases?

src/bindings.zig
```diff
@@ -0,0 +1,52 @@
+const std = @import("std");
```
slimsag, line 1: Two thoughts for improvement:

1. Should these bindings live here, or in `libmach` I wonder? The same question would go for other Mach libraries in the future, e.g. when we expose a C ABI for the ECS library, I wonder where it should live?

2. We should plan on changing this API to reflect what Ayush said in Matrix, something like this as our long-term goal (but may require some changes to make this possible):

```
int main(void) {
    mach_platform_init(init_fn, usr_ptr);
    while (...) {
        mach_platform_update(update_fn, resize_fn, usr_ptr);
    }
    mach_platform_deinit(deinit_fn, usr_ptr);
}
```

src/bindings.zig
```diff
@@ -0,0 +1,52 @@
...
+const gpu = @import("gpu");
+const Core = @import("Core.zig");
+const libmach = @import("platform/libmach.zig");
+const native = @import("platform/native.zig");
+
+// Current Limitations:
+// 1. Currently, ecs seems to be using some weird compile-time type trickery, so I'm not exactly sure how
+// `engine` should be integrated into the C API
+// 2. Core might need to expose more state so more API functions can be exposed (for example, the WebGPU API)
```
slimsag, line 10: Once https://github.com/hexops/mach/pull/403 lands we should end up being ABI-compatible with the `webgpu.h` API, and Zig's stage2 compiler could even emit `webgpu.h` for us based on this too I think. Should be possible to start writing Mach core / WebGPU apps using this.

libmach/test.lisp
```diff
@@ -0,0 +1,64 @@
+;; Tests the behavior of `libmach` using Common Lisp's CFFI
```
slimsag, line 1: Just curious @zack466 which language do you actually plan to use this from? I've personally never used lisp, but would be interested in maintaining some official Go bindings potentially.

src/bindings.zig
iddev5: My opinion on (1) I think the bindings should live inside src/ maybe changing the name to libmach.zig to be more clear. The libmach directory currently doesn't serves any purpose, it just contains examples.

libmach/test.lisp
zack466: Common Lisp, haha

CFFI code is dead simple in Lisp (as you can see in `test.lisp`), which is really nice. I was actually originally trying to work with bgfx, but it was a huge pain to compile and use properly, so I eventually gave up. With Mach, on the other hand, a simple `zig build` does the trick!

PR #411, opened 2022-07-14
sysaudio: Initial WebAudio backend implementation

iddev5:
It can currently request devices (audio contexes) and run them, though data can't be processed. Device iterator also doesn't work yet.

Makes the necessary build.zig changes for examples to run.


sysaudio/src/webaudio.zig
```diff
@@ -0,0 +1,108 @@
...
+}
+
+pub fn waitEvents(_: Audio) void {}
+
+const default_channel_count = 2;
+const default_sample_rate = 48000;
+
+pub fn requestDevice(audio: Audio, config: DeviceDescriptor) Error!Device {
+    // NOTE: WebAudio only supports F32 audio format, so config.format is unused
+    const mode = config.mode orelse .output;
```
alichraghi, line 60: shouldn't we require `mode` (like desktop implementation)? i don't think it's hard for user to just do `.{ .mode = .output }`. also has readibility benefit

slimsag: We should change `DeviceDescriptor`'s `mode` field:

```diff
-mode: ?Mode = null,
+mode: Mode = .output,
```

slimsag (approved):
Super exciting!

PR #416, opened 2022-07-18
freetype: fix tests for stage2 and Face.getGlyphName()

PiergiorgioZagaria:

The harfbuzz tests randomly failed to compile with stage2. With the usual fix it doesn't fail anymore, but I wonder how it managed to compile before, and since the behaviour is random there must be some UB in the compiler regarding this. Should I write about this in the stage2 issue?

Also `Face.getGlyphName()` returned a slice from a buffer on the Stack. I decided to pass the slice to the function, so that memory management of the buffer is left to the user, though this means that the user still has to call `sliceTo()` on his buffer to get the actual string.

### DIFF
--- a/freetype/src/freetype/Face.zig
+++ b/freetype/src/freetype/Face.zig
@@ -137,10 +137,8 @@ pub fn getTrackKerning(self: Face, point_size: i32, degree: i32) Error!i32 {
     return kerning;
 }
 
-pub fn getGlyphName(self: Face, index: u32) Error![]const u8 {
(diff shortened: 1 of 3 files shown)


slimsag (approved):
Looks good, thanks for this.

> since the behaviour is random there must be some UB in the compiler regarding this. Should I write about this in the stage2 issue?

Assuming this works, I'm sure it's fine & the Zig authors will catch this on their own (if they haven't already), but feel free to write about it there if you like!

PR #420, opened 2022-07-19
libmach: update the C API - expose init, update, and deinit functions

zack466:
This pull request reworks the libmach C API.
- Instead of working with callbacks,`libmach` exposes functions such as `mach_init_core`, `mach_update`, and `mach_deinit` to be called by users of the shared library
- Examples using this new API can be found in the `libmach/` directory.


src/platform/libmach.zig
```diff
@@ -2,34 +2,106 @@ const std = @import("std");
...
 
-pub fn update(_: *App, core: *Core) !void {
-    core_callbacks.core_update.?(core);
+pub export fn mach_window_should_close(core: *Core) bool {
+    return core.internal.window.shouldClose();
+}
+
+pub const CoreCallback = fn (*Core, u32, u32) callconv(.C) void;
+
+// Adapted from native.zig
```
slimsag, line 62: I think to increase the odds that `libmach` is always up-to-date and working, and to get higher odds of it also working on other platforms (webassembly for example), we should instead have `native.zig` export an API which `libmach.zig` calls into. Then `libmach.zig` becomes just rather simple C export wrappers.

@iddev5 do you agree with that?

slimsag (approved):
A clear improvement over what we have right now, so merging!

src/platform/libmach.zig
iddev5: Yes, this was the idea all along. Plus these functions will be able to reduce the code duplication which we have with mach_update and things like that.

iddev5:
Good PR. I didnt got time to review it but one minor thing I would say is that namings are odd and inconsistent. I think we should just follow the ``mach_<module>_<function>`` convention. So that makes ``mach_init_core`` as ``mach_core_init``. Just a thought.

PR #423, opened 2022-07-20
libmach: update API again, factors out init/update/deinit from native

zack466:
This PR makes a few changes to the libmach API in accordance with the comments on the previous PR.

Also, it seems like I forgot to stage my changes to `build.zig` in the previous commit, so they are included in this commit instead.

### DIFF
--- a/libmach/test.c
+++ b/libmach/test.c
@@ -4,17 +4,18 @@
 
 typedef void resize_callback(void*, uint32_t, uint32_t);
 
-typedef enum {
-  MachFailure,
-  MachSuccess
-} MachReturn;
+typedef enum MachStatus {
+    MachStatus_Success = 0x00000000,
+    MachStatus_Error = 0x00000001,
+    MachStatus_Force32 = 0x7FFFFFFF
+} MachStatus;
 
 // `libmach` exported API bindings
 void* mach_core_init(void);
 void mach_core_deinit(void*);
 void mach_core_set_should_close(void*);
 bool mach_core_window_should_close(void*);
-MachReturn mach_core_update(void*, resize_callback);
+MachStatus mach_core_update(void*, resize_callback);
 float mach_core_delta_time(void*);
 
 void resize_fn(void* core, uint32_t width, uint32_t height) {
@@ -32,7 +33,7 @@ int main() {
   }
 
   while (!mach_core_window_should_close(core)) {
-    if (mach_core_update(core, resize_fn) == MachFailure) {
+    if (mach_core_update(core, resize_fn) == MachStatus_Error) {
       printf("Error updating Mach\n");
       break;
     };
--- a/libmach/test.c
+++ b/libmach/test.c
@@ -4,13 +4,18 @@
 
 typedef void resize_callback(void*, uint32_t, uint32_t);
 
+typedef enum {
+  MachFailure,
+  MachSuccess
+} MachReturn;
+
 // `libmach` exported API bindings
-void* mach_init_core(void);
-void mach_deinit(void*);
-void mach_set_should_close(void*);
-bool mach_window_should_close(void*);
-int mach_update(void*, resize_callback);
-float mach_delta_time(void*);
+void* mach_core_init(void);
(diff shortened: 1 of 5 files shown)


libmach/test.c
```diff
@@ -4,13 +4,18 @@
 
 typedef void resize_callback(void*, uint32_t, uint32_t);
 
+typedef enum {
+  MachFailure,
+  MachSuccess
+} MachReturn;
```
slimsag, line 10: I wonder if it might be a good idea to have these prefixed with an underscore, use a `_Force32` to force the enum to be 32-bits (in case someone puts it somewhere expecting a fixed size), similar to how [webgpu.h does it](https://github.com/webgpu-native/webgpu-headers/blob/64db862b70cee7a5a6dc89dc7d920a0300125f4a/webgpu.h#L150-L158). We could also name this "Status" instead of "Return" so that there could be other error status codes in the future:

```c
typedef enum MachStatus {
    MachStatus_Success = 0x00000000,
    MachStatus_Error = 0x00000001,
    MachStatus_Force32 = 0x7FFFFFFF
} MachStatus;
```

Maybe you have a different idea for error handling though?

src/platform/native.zig
```diff
@@ -587,60 +587,81 @@ pub const Platform = struct {
...
+
+        try app.update(core);
+    }
+}
+
+pub fn core_init(allocator: std.mem.Allocator) !*Core {
+    const core: *Core = try allocator.create(Core);
+    core.* = try Core.init(allocator);
+
+    // // Glfw specific: initialize the user pointer used in callbacks
```
slimsag, line 614: should remove double comment here

src/platform/native.zig
```diff
@@ -587,60 +587,81 @@ pub const Platform = struct {
...
+    core.*.internal.initCallback();
+
+    return core;
+}
+
+pub export fn core_deinit(core: *Core) void {
```
slimsag, line 620: I think this function shouldn't be exported.

Also, while `snake_case` is correct in C (and `libmach.zig`), here it is not. The function names in this file should be `coreDeinit`, `coreInit`, `coreUpdate` - quoting the [Zig style guide](https://ziglang.org/documentation/master/#toc-Names):

> Roughly speaking: camelCaseFunctionName, TitleCaseTypeName, snake_case_variable_name. More precisely: 

Of course what you did in `libmach.zig` is correct, since those names must match C conventions as exported functions.

src/platform/native.zig
```diff
@@ -587,60 +587,81 @@ pub const Platform = struct {
...
+    core.internal.deinit();
+    // assumes that core.allocator is the same allocator used to allocate core
+    core.allocator.destroy(core); // "I used the core to destroy the core"
```
slimsag, line 623: How about we pass the allocator into this function instead? Would eliminate these semi-confusing comments, and is a better approach to use generally anyway.

iddev5 (commented):
This is a single review of the overall PR. While the PR implements everything well in its own ways, I think some things are confusing.

- Core is an actual module in mach. This means all C functions with ``mach_core`` should map directly to src/Core.zig
- What we are currently doing is a bit of magic. We are actually mixing parts from Platform (src/platform/native.zig) and Core(src/core.zig). This can cause confusion to people who are looking at the zig API for docs, because in the long run thats what people would do -- try to replicate the zig examples in C comparing the function names.
- I think both should be kept separate. Core itself can be perfectly used without needing the mach entry point, and I think libmach should respect it. What in my suggestion implies separating the current API into to ``mach_platform_*`` and ``mach_core_*``.

I m not suggesting to do it now with this PR. Doing it slowly over time is fine. I m just suggesting that there should be some goal/rules in order to clear up the path ahead. Thoughts?

libmach/test.c
zack466: I'm not exactly an expert on error handling in C... just having error codes is probably fine for now.

zack466:
Alright, I've made a few changes

slimsag (approved):

PR #425, opened 2022-07-20
all: use refAllDeclsRecursive to find and fix mistakes

alichraghi:
sysjs tests still fail because js functions aren't linked. cc @iddev5 may have a fix for this

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -18,16 +18,24 @@ pub fn build(b: *std.build.Builder) void {
     };
     const options = Options{ .gpu_dawn_options = gpu_dawn_options };
 
-    // TODO: re-enable tests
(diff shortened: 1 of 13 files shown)


iddev5:
Sysjs cant be tested normally because its a wasm application. My suggestion is to just remove it from testing for now. Actually we need to design a different kind of test runner specifically for wasm, which is quite a lot of work on its own and is out of scope of this PR.

iddev5:
Good catch on sysjs integer type mistakes, not sure why it never appeared on me.

slimsag (approved):

alichraghi:
oops, didn't know that was a thing

PR #429, opened 2022-07-23
CI: fixes

alichraghi:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,11 +18,11 @@ jobs:
         uses: actions/checkout@v2
       - name: Setup Zig
         run: |
-          sudo apt install xz-utils
+          sudo apt install xz-utils libasound2-dev
           sudo sh -c 'wget -c https://ziglang.org/builds/zig-linux-x86_64-0.10.0-dev.3027+0e26c6149.tar.xz -O - | tar -xJ --strip-components=1 -C /usr/local/bin'
       - name: x86_64-linux -> x86_64-linux-musl
         if: matrix.project == 'freetype' || matrix.project == 'glfw'
-        run: cd ${{ matrix.project }} && zig build -Dtarget=x86_64-linux-mul
+        run: cd ${{ matrix.project }} && zig build -Dtarget=x86_64-linux-musl
       - name: x86_64-linux -> x86_64-macos
         run: cd ${{ matrix.project }} && zig build -Dtarget=x86_64-macos.12-none
         env:
@@ -63,10 +63,10 @@ jobs:
       - name: Setup Zig
         run: |
           $ProgressPreference = 'SilentlyContinue'
-          Invoke-WebRequest -Uri "https://ziglang.org/builds/zig-windows-x86_64-0.10.0-dev.3027+0e26c6149.zip" -OutFile "C:\zig.zip"
+          Invoke-WebRequest -Uri 'https://ziglang.org/builds/zig-windows-x86_64-0.10.0-dev.3027+0e26c6149.zip' -OutFile 'C:\zig.zip'
           cd C:\
           7z x zig.zip
-          Add-Content $env:GITHUB_PATH "C:\zig-windows-x86_64-0.10.0-dev.3027+0e26c6149\"
+          Add-Content $env:GITHUB_PATH 'C:\zig-windows-x86_64-0.10.0-dev.3027+0e26c6149\'
       - name: compile all examples
         if: matrix.project == '.'
         run: cd ${{ matrix.project }} && zig build compile-all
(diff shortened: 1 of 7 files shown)


freetype/build.zig
```diff
@@ -48,6 +48,12 @@ pub fn build(b: *std.build.Builder) !void {
     const mode = b.standardReleaseOptions();
     const target = b.standardTargetOptions(.{});
 
+    const test_app = b.addStaticLibrary("test_app", null);
+    test_app.setBuildMode(mode);
+    test_app.setTarget(target);
+    link(b, test_app, .{ .freetype = .{ .brotli = true } });
+    test_app.install();
```
slimsag, line 55: I've been making an effort to not create unused static libraries (which add to compile times), and not to install them for the reasons described in https://github.com/hexops/mach/issues/232

This change and the GLFW change below go against that. But why? How's it helpful?

alichraghi: because `zig build` need to do something. otherwise e.g cross-compiling in CI is pointless

.github/workflows/ci.yml
```diff
@@ -7,22 +7,22 @@ jobs:
...
     env:
-      DISPLAY: ':99.0'
+      DISPLAY: ":99.0"
     steps:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Setup Zig
         run: |
-          sudo apt install xz-utils
+          sudo apt install xz-utils libasound2-dev
```
slimsag, line 21: Either this should be just `libasound2` (not `-dev`), or whatever header dependency we have in here should be distributed via https://github.com/hexops/sdk-linux-x86_64 and https://github.com/hexops/sdk-linux-aarch64 instead.

.github/workflows/ci.yml
```diff
@@ -7,22 +7,22 @@ jobs:
...
     strategy:
       matrix:
-        project: ['.', 'glfw', 'freetype']
+        project: [".", "glfw", "freetype"]
     # We want to run on external PRs, but not on our own internal PRs as they'll be run by the push
     # to the branch.
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     env:
-      DISPLAY: ':99.0'
+      DISPLAY: ":99.0"
```
slimsag, line 15: It's a bit unfortunate we now have different formatting for CI files throughout this repository.

I suggest we add a new script: `dev/ci-lint.sh` which runs:

* actionlint: https://github.com/rhysd/actionlint/blob/main/docs/install.md
* Some YAML formatter.. preferable not one from the npm ecosystem, maybe this Go one https://github.com/UltiRequiem/yamlfmt

on all our CI yml files.

freetype/build.zig
slimsag: OK, I see what you're going for here. But it doesn't seem like we should `.install()` these, as they are for testing only after all.

Separately, I want to ask: does `zig build` not build the test binary themselves? It seems to me we should be able to have `zig build` build the tests.. and those should build for the target OS, too - and then we wouldn't need to build useless static libraries like this.

.github/workflows/ci.yml
slimsag: (I can address this in a future PR)

.github/workflows/ci.yml
slimsag: (should address this before merging)

freetype/build.zig
slimsag: (should address this before merging)

alichraghi: there's `b.testExe`

.github/workflows/ci.yml
alichraghi: i'd tried without `-dev` but it didn't work. i do the same in my machine (linux mint)

.github/workflows/ci.yml
alichraghi: `yamlfmt` results the same output. at least im gonna say it has nothing to do with quotes `"`, `'`

freetype/build.zig
slimsag: `addTestExe` looks good to me!

.github/workflows/ci.yml
slimsag: What part about it didn't work? This would indicate cross-compilation won't work

.github/workflows/ci.yml
slimsag: To be clear, I am fine with the formatting being anything (double quotes, whatever) as long as we have a way for it to be consistent. We can use yamlfmt.

.github/workflows/ci.yml
alichraghi: missing headers/static libs

.github/workflows/ci.yml
alichraghi: i think **actionlint** is enough and yaml formatters doesn't make any change as long as we just edit github workflows. that's why there's no many yaml formatters, or at least they are unused in most cases

slimsag: https://github.com/hexops/mach/issues/430

.github/workflows/ci.yml
slimsag: https://github.com/hexops/mach/issues/431

freetype/build.zig
slimsag: https://github.com/hexops/mach/pull/429/files#r928133589

slimsag (approved):
Filed issues for my outstanding concerns with this PR. Will merge now since there are good changes here otherwise.

alichraghi:
omg i had a lot of changes. forgot to mark is draft lol

slimsag:
ah gotcha, my bad

PR #434, opened 2022-07-24
all: build: install tests exe

### LINKED ISSUE #433: {freetype,glfw}: should not build and install useless static lib just for testing
https://github.com/hexops/mach/pull/429/files#r928133589

alichraghi:
Closes #433 


build.zig
```diff
@@ -18,25 +18,27 @@ pub fn build(b: *std.build.Builder) void {
...
     const options = Options{ .gpu_dawn_options = gpu_dawn_options };
 
-    const main_tests = b.addTest("src/main.zig");
+    const main_tests = b.addTestExe("mach-tests", "src/main.zig");
     main_tests.setBuildMode(mode);
     main_tests.setTarget(target);
     main_tests.addPackage(pkg);
     main_tests.addPackage(gpu.pkg);
     main_tests.addPackage(glfw.pkg);
+    main_tests.install();
```
slimsag, line 27: Is it actually right to _install_ tests? This would place them in the `zig-out` directory as I understand it, which feels not ideal to me. For example, if we were to begin publishing binary releases I don't think we would want these included.

slimsag (approved):
LGTM excluding the installation point

build.zig
alichraghi: i think we should just wait for zig to provide a way to compile without producing binaries

slimsag (approved):

PR #435, opened 2022-07-26
 dev: add other workflows to ci-lint.sh

### LINKED ISSUE #432: dev: ci-lint.sh should run on all actions

alichraghi:
Fixes #432

### DIFF
--- a/dev/ci-lint.sh
+++ b/dev/ci-lint.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
-# actionlint: https://github.com/rhysd/actionlint
-actionlint .github/workflows/ci.yml .github/workflows/m1_ci.yml
+set -exuo pipefail
+cd "$(dirname "${BASH_SOURCE[0]}")"/..
 
+# actionlint: https://github.com/rhysd/actionlint
+./actionlint $(find . | grep -v '\.git/' | grep '\.github/workflows/' | grep '.yml')


dev/ci-lint.sh
```diff
@@ -1,4 +1,8 @@
 #!/usr/bin/env bash
 # actionlint: https://github.com/rhysd/actionlint
-actionlint .github/workflows/ci.yml .github/workflows/m1_ci.yml
+actionlint .github/workflows/ci.yml .github/workflows/m1_ci.yml \
+	gpu-dawn/.github/workflows/draft_release.yml .github/workflows/push_subrepos.yml \
+	freetype/.github/workflows/ci.yml glfw/.github/workflows/ci.yml \
+	glfw/.github/workflows/m1_ci.yml gpu-dawn/.github/workflows/publish_release.yml \
+	gpu-dawn/.github/workflows/ci.yml gpu-dawn/.github/workflows/m1_ci.yml
```
slimsag, line 7: Can we use a `find` command to find these files instead of manually listing them all out? Otherwise this will quickly become out of date

slimsag: Perhaps:

```sh
actionlint $(find .. | grep -v '\.git/' | grep '\.github/workflows/' | grep '.yml')
```

slimsag: We should also change this script to use this at the top, so it fails if any commands in it fail, and so we always run from the repo root:

```
#!/usr/bin/env bash
set -exuo pipefail
cd "$(dirname "${BASH_SOURCE[0]}")"/..
```

slimsag (approved):

Issue #405, opened 2022-07-12
build: types and packages are mess


alichraghi:
the goal of this issue is try to find out a way to not repeat codes in sub-projects and also build files
for example same `Options` struct exist in `./build.zig` and `gpu/build.zig`. or `refAllDecls` function (which is very useful for tests and we should consider to use it in other packages) has been declared twice in **freetype** and **audio/soundio**

PiergiorgioZagaria:
I think the reason for the code repetition was because we can't import from `../`. For example we could have a `build_utils.zig` file with all the structs and functions commonly used in the `build.zig` files, but we wouldn't be able to import it.

Also, there is the fact that we are offering people the ability to use mach packages by themselves without adding all of mach. For example if someone wanted to use just `mach-glfw` it would still need that `build_utils.zig`.

alichraghi:
> Have any issues been caused by this? 

when hacking on #396, i had to use `@bitCast` while passing the exact same `Options` struct to `gpu.testStep`
Mach is definitely going to get bigger, and with the increase of such problems, controlling  code quality will become more difficult

slimsag:
I agree those bitcasts between package parameters are an issue 👍 But I think they're caused by us using symlinks to load packages, not because of code duplication

alichraghi:
i think this can be closed as of https://github.com/hexops/mach/commit/fe4e721508ac7990b17d12c7f0c0f25a58f39d8d

PR #437, opened 2022-07-27
gamemode: Implement FeralInteractive/gamemode support

PiergiorgioZagaria:

The previous comment assumed that we needed to compile the gamemode shared library ourselves, which is wrong, so I deleted the old comment.

As a sidenote: `getEnvVarOwned()` was taken from utils.zig (it is used to check GPU_BACKEND), when exiting the program, the gpa said there were memory leaks, which were caused by `getEnvVarOwned()`. Should we free the other strings where it's called too?


src/platform/native.zig
```diff
@@ -214,6 +238,23 @@ pub const Platform = struct {
         while (platform.events.popFirst()) |ev| {
             platform.allocator.destroy(ev);
         }
+
+        if (builtin.os.tag == .linux) {
```
slimsag, line 242: Maybe move these blocks into a separate function we call from here, e.g. `platform.deinitLinuxGameMode()`

src/platform/native.zig
```diff
@@ -44,7 +45,30 @@ pub const Platform = struct {
...
     };
 
+    fn getEnvVarOwned(allocator: std.mem.Allocator, key: []const u8) error{ OutOfMemory, InvalidUtf8 }!?[]u8 {
+        return std.process.getEnvVarOwned(allocator, key) catch |err| switch (err) {
+            error.EnvironmentVariableNotFound => @as(?[]u8, null),
+            else => |e| e,
+        };
+    }
     pub fn init(allocator: std.mem.Allocator, core: *Core) !Platform {
+        if (builtin.os.tag == .linux) {
```
slimsag, line 55: Same here, so the `init` function doesn't grow too large.

gamemode-zig/gamemode.zig
```diff
@@ -0,0 +1,76 @@
+//! Note: All the above requests can be blocking - dbus requests can and will block while the daemon
+//! handles the request. It is not recommended to make these calls in performance critical code
+
+pub const c = @import("gamemode_client.zig");
```
slimsag: what's the motivation for using translate-c vs. just directly importing the header?

slimsag:
> As a sidenote: getEnvVarOwned() was taken from utils.zig (it is used to check GPU_BACKEND), when exiting the program, the gpa said there were memory leaks, which were caused by getEnvVarOwned(). Should we free the other strings where it's called too?

Definitely, it probably was copied from a build script where free'ing is not important due to arena allocator. Do you want to include a commit to fix it (or send another PR)?

gamemode-zig/gamemode.zig
PiergiorgioZagaria: `zig translate-c` produced a zig file with an error that made it not compile. It tried to `@alignCast(std.meta.alignment(?*anyopaque), &REAL_internal_gamemode_request_start)`, for all the functions. In fetch_upstream.sh I delete that part to make it compile

PiergiorgioZagaria:
After checking more closely I noticed the `getEnvVarOwned()` that leaked were all in the build scripts, I guess there is no need to fix them

build.zig
```diff
@@ -1,8 +1,8 @@
 const std = @import("std");
 const builtin = @import("builtin");
 pub const gpu = @import("gpu/build.zig");
-const gpu_dawn = @import("gpu-dawn/build.zig");
-pub const glfw = @import("glfw/build.zig");
+const gpu_dawn = @import("gpu/libs/mach-gpu-dawn/build.zig");
+pub const glfw = @import("gpu/libs/mach-glfw/build.zig");
```
slimsag, line 5: did these symlink changes make it into here by accident?

PiergiorgioZagaria: I was testing if it worked and forgot to go back to how it was before

src/platform/native.zig
```diff
@@ -214,6 +223,45 @@ pub const Platform = struct {
...
+    fn initLinuxGamemode(allocator: std.mem.Allocator) error{ OutOfMemory, InvalidUtf8 }!void {
+        if (builtin.os.tag == .linux) {
+            const gamemode = @import("gamemode");
+            const GAMEMODE_ENV = try getEnvVarOwned(allocator, "GAMEMODE");
+            if (GAMEMODE_ENV == null) {
+                gamemode.requestStart() catch |err| {
+                    std.log.err("Gamemode error {} -> {s}", .{ err, gamemode.errorString() });
+                };
+            } else {
+                if (!std.ascii.eqlIgnoreCase(GAMEMODE_ENV.?, "off")) {
```
slimsag, line 239: Maybe check for both `off` and `false` here?

slimsag: We could store whether or not we used gamemode in a field of `platform` struct, so that we can reuse it simply in the deinit code without duplicating the env var handling logic

gamemode-zig/gamemode.zig
slimsag: Ahh got it, makes sense

build.zig
alichraghi: hopefully this fixes #405

slimsag:
LGTM once https://github.com/hexops/mach/pull/437#discussion_r932012343 is addressed

slimsag (approved):
🔥 you rock, really excited to have this!

PR #442, opened 2022-07-31
glfw: add shared option which builds glfw into it's own shared library

kcbanner:
This adds the `shared` option, which builds GLFW into it's own shared library. 

The use case for this is as follows (in this setup, the exe loads the game dll at runtime using std.DynLib):
```
    // Engine launcher
    const exe = b.addExecutable("ion", "src/main.zig");
    {
         ... setup the exe build step ...
        exe.addPackage(glfw.pkg);
        glfw.link(b, exe, .{
            .shared = true
        });
    }

    // Game dll
    const game = b.addSharedLibrary("game", "game/src/game.zig", .unversioned);
    {
        ... setup the game build step ...
        game.addPackage(glfw.pkg);
        glfw.link(b, game, .{
            .shared = true
        });
    }
```

This way, both `game` and `exe` link with the shared library, so that there is one copy of the `_glfw` global state in the process. Static linking wouldn't work here, as the dll and exe would have separate copies of `_glfw`.

A couple things I'm not sure about:
- I opted to cache the shared library for each call (create it on the first `glfw.link` call and re-used the instance across calls). The other option would be a separate call (ie. `glfw.createSharedGlfw`) to create the shared lib, and you'd pass *that* as the option instead of a bool. I'm happy to change over to that, but I was trying to keep the API simple.
- Specifying `zig-out/bin` as the output dir - this works in the simple case, but may break if that directory has been changed on the parent step. The intention is to place the dll next to whatever exe/dll is linking with it, but I'm not familiar enough with the build system to know how to get that automatically.

### DIFF
--- a/glfw/build.zig
+++ b/glfw/build.zig
@@ -8,14 +8,15 @@ pub fn build(b: *Builder) void {
     const target = b.standardTargetOptions(.{});
 
     const test_step = b.step("test", "Run library tests");
-    test_step.dependOn(&testStep(b, mode, target).step);
+    test_step.dependOn(&testStep(b, mode, target, "glfw_tests", .{}).step);
+    test_step.dependOn(&testStep(b, mode, target, "glfw_shared_tests", .{ .shared = true }).step);
 }
 
-pub fn testStep(b: *Builder, mode: std.builtin.Mode, target: std.zig.CrossTarget) *std.build.RunStep {
-    const main_tests = b.addTestExe("glfw_tests", thisDir() ++ "/src/main.zig");
+pub fn testStep(b: *Builder, mode: std.builtin.Mode, target: std.zig.CrossTarget, exe_name: []const u8, options: Options) *std.build.RunStep {
+    const main_tests = b.addTestExe(exe_name, thisDir() ++ "/src/main.zig");
     main_tests.setBuildMode(mode);
     main_tests.setTarget(target);
-    link(b, main_tests, .{});
+    link(b, main_tests, options);
     main_tests.install();
     return main_tests.run();
 }
@@ -46,6 +47,9 @@ pub const Options = struct {
 
     /// System SDK options.
     system_sdk: system_sdk.Options = .{},
+
+    /// Build and link GLFW as a shared library.
+    shared: bool = false,
 };
 
 pub const pkg = std.build.Pkg{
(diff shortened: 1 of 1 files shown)


glfw/build.zig
```diff
@@ -8,14 +8,15 @@ pub fn build(b: *Builder) void {
...
 
     const test_step = b.step("test", "Run library tests");
-    test_step.dependOn(&testStep(b, mode, target).step);
+    test_step.dependOn(&testStep(b, mode, target, "glfw_tests", .{}).step);
+    test_step.dependOn(&testStep(b, mode, target, "glfw_shared_tests", .{ .shared = true }).step);
 }
 
-pub fn testStep(b: *Builder, mode: std.builtin.Mode, target: std.zig.CrossTarget) *std.build.RunStep {
-    const main_tests = b.addTestExe("glfw_tests", thisDir() ++ "/src/main.zig");
+pub fn testStep(b: *Builder, mode: std.builtin.Mode, target: std.zig.CrossTarget, exe_name: []const u8, options: Options) *std.build.RunStep {
```
slimsag, line 15: Could you leave this function signature as it was before, so that e.g. [this does not break](https://github.com/hexops/mach/blob/e6d7faaabab303d006ea998547c6dc7ce2305c2a/build.zig#L33) - and instead introduce a new private `fn testStepShared` & just duplicate the code here to test the shared library approach?

glfw/build.zig
```diff
@@ -8,14 +8,15 @@ pub fn build(b: *Builder) void {
...
+    const main_tests = b.addTestExe(exe_name, thisDir() ++ "/src/main.zig");
```
alichraghi, line 16: i think this should be applied to all sub-projects

glfw/build.zig
alichraghi: i think this is a good approach and should be applied to all sub-projects

slimsag: not opposed to that, let's do it in another PR though.

slimsag:
@kcbanner I pushed a few changes / cleanups, give this a try and let me know if it still works for you?

I also removed the line that was writing the lib into the bin dir, instead opting for the natural Zig output location so we end up with e.g.:

```
zig-out
zig-out/bin
zig-out/bin/glfw_tests
zig-out/bin/glfw_tests_shared
zig-out/lib
zig-out/lib/libglfw.dylib
```

PR #465, opened 2022-08-20
glfw: stage2 fn pointer update

Vulfox:

Saw some stage2 updates, figured I would throw in the glfw updates I needed to make for Vulkan to work. If you already have it locally and just making a bunch of other updates with it, feel free to shoot this PR down.


slimsag:
Ah sorry I hadn't seen this sooner or I would've merged it :) This is fixed already, thanks for sending the PR though!

PR #467, opened 2022-08-22
shaderexp: Use fullscreen triangle instead of quad, move vertex data into the shader

praschke:
Using a fullscreen triangle is (very) marginally better than a quad because it avoids overdraw and isn't any more complicated. Moving the static vertex data for the fullscreen primitive into the shader simplifies command buffer recording and pipeline layout.


slimsag (approved):
Thanks for correcting this!

PR #468, opened 2022-08-22
examples: fix old example assets directory paths

LordMZTE:
This commit fixes a bug where the build script would look for
an assets directory that has been moved.


slimsag (approved):
Thanks for fixing this!

PR #476, opened 2022-08-25
Gamemode

alichraghi:

### DIFF
--- a/gamemode/gamemode_client.zig
+++ b/gamemode/gamemode_client.zig
@@ -532,9 +532,9 @@ pub var internal_gamemode_client_error_string: [512]u8 = [1]u8{
     0,
 } ++ [1]u8{0} ** 511;
 pub var internal_libgamemode_loaded: c_int = 1;
-pub const api_call_return_int = ?fn () callconv(.C) c_int;
-pub const api_call_return_cstring = ?fn () callconv(.C) [*c]const u8;
-pub const api_call_pid_return_int = ?fn (pid_t) callconv(.C) c_int;
+pub const api_call_return_int = ?*const fn () callconv(.C) c_int;
+pub const api_call_return_cstring = ?*const fn () callconv(.C) [*c]const u8;
+pub const api_call_pid_return_int = ?*const fn (pid_t) callconv(.C) c_int;
 pub var REAL_internal_gamemode_request_start: api_call_return_int = null;
 pub var REAL_internal_gamemode_request_end: api_call_return_int = null;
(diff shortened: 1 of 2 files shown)


src/platform/native.zig
```diff
@@ -204,7 +204,9 @@ pub const Platform = struct {
         if (builtin.os.tag == .linux and platform.linux_gamemode_is_active) {
             const gamemode = @import("gamemode");
             gamemode.requestEnd() catch |err| {
-                std.log.err("Gamemode error {} -> {s}", .{ err, gamemode.errorString() });
+                if (!std.mem.containsAtLeast(u8, gamemode.errorString(), 1, "dlopen failed")) {
```
slimsag, line 207: Maybe add comment here:

```
// We don't want these warnings to appear if you don't have gamemode installed, so do a simple string check:
//
// error: Gamemode error error.RequestFailed -> dlopen failed - libgamemode.so: cannot open shared object file: No such file or directory
```

But also, is this the right place for this actually ? I'm surprised it is in the `deinitLinuxGamemode` function and not `initLinuxGamemode` that causes this error to get printed

gamemode/gamemode_client.zig
```diff
@@ -532,9 +532,9 @@ pub var internal_gamemode_client_error_string: [512]u8 = [1]u8{
     0,
 } ++ [1]u8{0} ** 511;
 pub var internal_libgamemode_loaded: c_int = 1;
-pub const api_call_return_int = ?fn () callconv(.C) c_int;
-pub const api_call_return_cstring = ?fn () callconv(.C) [*c]const u8;
-pub const api_call_pid_return_int = ?fn (pid_t) callconv(.C) c_int;
+pub const api_call_return_int = ?*const fn () callconv(.C) c_int;
+pub const api_call_return_cstring = ?*const fn () callconv(.C) [*c]const u8;
+pub const api_call_pid_return_int = ?*const fn (pid_t) callconv(.C) c_int;
```
slimsag, line 537: Until the Zig self-hosted compiler stabilizes a bit, I am suggesting others still use the slightly older Zig version: https://github.com/hexops/mach#supported-zig-version

This change would prevent Mach from building on the older version, though. Could you change this PR to have both versions of this file, a `gamemode_client_stage1.zig` and a `gamemode_client.zig` and choose the right one depending on compiler?

alichraghi:
i convert this to draft. because i can't make sure this is working correctly until cImport issue solved

PR #487, opened 2022-08-28
build: fixes

alichraghi:


alichraghi:
Closes #490

libs/ecs/build.zig
```diff
@@ -9,6 +9,10 @@ pub fn build(b: *std.build.Builder) void {
 
 pub fn testStep(b: *std.build.Builder, mode: std.builtin.Mode, target: std.zig.CrossTarget) *std.build.RunStep {
     const main_tests = b.addTestExe("ecs-tests", (comptime thisDir()) ++ "/src/main.zig");
+
+    // TODO: remove this when tests passed with self-hosted compiler
```
slimsag, line 13:
```suggestion
    // TODO(self-hosted): remove this when tests passed with self-hosted compiler
```

slimsag (approved):

mitchellh:
My only recommendation on this is out of an abundance of caution, I'd include at LEAST a `@sizeOf` test in these structs to ensure they're never broken. Its easy to do with padding miscalculations.

This will get better in stage2 with explicit integer-backed packed structs...

alichraghi:
Don't worry. I did tested all structs carefully (not just size).
Also in case of invalid size, library won't compile. Because all flags are passed to functions, that expects a valid size.

PR #514, opened 2022-09-04
gpu-dawn: default to release version of Dawn

### LINKED ISSUE #464: gpu-dawn: default to release version of Dawn by default (even in debug builds)
The nanovg Mach backend by dweiller on Matrix is ~15FPS in a debug build, and much much faster in a release build. Granted, it's not using WebGPU optimally, but it goes to show the impact many WebGPU API calls can have in a debug vs. release build.

Dawn already has a fair amount of debugging/validation at runtime, since after all it is used in Chrome as a release build not a debug build, so we should default to using a release build of Dawn even when building a debug Zig application.

Similar to how we have `-Ddawn-from-source` we can have `-Ddawn-debug` to use a debug build, though.

TL;DR:

* Better performance
* Not much worse debugging (Chrome uses release builds after all)
* Less potential UBSAN issues (e.g. these are causing Windows examples to fail right now)
* May improve linker times(?)

praschke:
possibly not the best way to fix #464, but straightforward and it stopped filling up my storage.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -27,6 +27,7 @@ pub fn build(b: *std.build.Builder) void {
 
     const gpu_dawn_options = gpu_dawn.Options{
         .from_source = b.option(bool, "dawn-from-source", "Build Dawn from source") orelse false,
+        .debug = b.option(bool, "dawn-debug", "Use a debug build of Dawn") orelse false,
     };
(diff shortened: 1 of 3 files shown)


slimsag (approved):
Good enough for now, thank you!

PR #518, opened 2022-09-09
{sysaudio,examples}: get sysaudio running on linux, separate audio configuration and descriptor

desttinghim:
This started with the sysaudio example running, but not showing up on my screen. @leroycep gave me a patch that made it show up by clearing the screen. 

My system is using pipewire running at a sample rate of 48000 hertz, so I was getting a lot of clicking from buffer underruns. I haven't yet fixed that because I got distracted by separating the configuration from the descriptor.

### DIFF
--- a/examples/sysaudio/main.zig
+++ b/examples/sysaudio/main.zig
@@ -67,6 +67,7 @@ pub fn update(app: *App, engine: *mach.Core) !void {
 // out using 1-log10(x*10) (google it to see how it looks, it's strong for most of the duration of
 // the note then fades out slowly.)
 pub const ToneEngine = struct {
+    sample_rate: f32 = 44100,
     playing: [2048]Tone = std.mem.zeroes([2048]Tone),
 
     const Tone = struct {
@@ -78,9 +79,20 @@ pub const ToneEngine = struct {
     pub fn render(engine: *ToneEngine, descriptor: sysaudio.DeviceDescriptor, buffer: []u8) void {
         // TODO(sysaudio): demonstrate how to properly handle format of the buffer here.
         // Right now we blindly assume f32 format, which is wrong (but always right in WASM.)
-        const sample_rate = @intToFloat(f32, descriptor.sample_rate.?);
-        const buf = @ptrCast([*]f32, @alignCast(@alignOf(f32), buffer.ptr))[0 .. buffer.len / @sizeOf(f32)];
-        const frames = buf.len / descriptor.channels.?;
+        engine.sample_rate = @intToFloat(f32, descriptor.sample_rate);
+        switch (descriptor.format) {
+            .U8 => renderWithType(u8, engine, descriptor, buffer),
+            .S16 => renderWithType(i16, engine, descriptor, buffer),
+            .S24 => renderWithType(i24, engine, descriptor, buffer),
+            .S32 => renderWithType(i32, engine, descriptor, buffer),
+            .F32 => renderWithType(f32, engine, descriptor, buffer),
+        }
+    }
+
+    pub fn renderWithType(comptime T: type, engine: *ToneEngine, descriptor: sysaudio.DeviceDescriptor, buffer: []u8) void {
+        const sample_rate = engine.sample_rate;
+        const buf = @ptrCast([*]T, @alignCast(@alignOf(T), buffer.ptr))[0 .. buffer.len / @sizeOf(T)];
+        const frames = buf.len / descriptor.channels;
 
         var frame: usize = 0;
         while (frame < frames) : (frame += 1) {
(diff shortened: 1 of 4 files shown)


examples/sysaudio/main.zig
```diff
@@ -40,12 +41,17 @@ pub fn update(app: *App, engine: *mach.Core) !void {
     while (engine.pollEvent()) |event| {
         switch (event) {
             .key_press => |ev| {
-                try app.device.start();
+                if (builtin.cpu.arch == .wasm32) try app.device.start();
```
slimsag, line 44: This isn't a wasm-specific API, it should be called here. What issue do you experience if it is?

examples/sysaudio/main.zig
```diff
@@ -61,6 +67,7 @@ pub fn update(app: *App, engine: *mach.Core) !void {
 // out using 1-log10(x*10) (google it to see how it looks, it's strong for most of the duration of
 // the note then fades out slowly.)
 pub const ToneEngine = struct {
+    sample_rate: f32 = 44100,
```
slimsag, line 70: This shouldn't be configured here / like this, it should come from the device/descriptor as it did before.

examples/sysaudio/main.zig
```diff
@@ -72,9 +79,20 @@ pub const ToneEngine = struct {
...
+        switch (descriptor.format) {
+            .U8 => renderWithType(u8, engine, descriptor, buffer),
+            .S16 => renderWithType(i16, engine, descriptor, buffer),
+            .S24 => renderWithType(i24, engine, descriptor, buffer),
+            .S32 => renderWithType(i32, engine, descriptor, buffer),
+            .F32 => renderWithType(f32, engine, descriptor, buffer),
+        }
+    }
+
+    pub fn renderWithType(comptime T: type, engine: *ToneEngine, descriptor: sysaudio.DeviceDescriptor, buffer: []u8) void {
```
slimsag, line 92: I wonder why this function should this not be receiving a typed buffer? Seems odd to handle that inside the function rather than at the call site

examples/sysaudio/main.zig
desttinghim: An assertion is triggered in libsoundio's alsa backend:
```
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0

info: mach: found Vulkan backend on Integrated GPU adapter: Intel(R) Xe Graphics (TGL GT2), Intel open-source Mesa driver: Mesa 22.0.4

info: channels 2
info: sample_rate 48000

example-sysaudio: /home/desttinghim/Code/mach/libs/sysaudio/upstream/soundio/src/alsa.c:1438: int outstream_start_alsa(struct SoundIoPrivate *, struct SoundIoOutStreamPrivate *): Assertion `!osa->thread' failed.
The following command terminated unexpectedly:
cd /home/desttinghim/Code/mach && /home/desttinghim/Code/mach/zig-out/bin/example-sysaudio
error: the following build command failed with exit code 1:
/home/desttinghim/Code/mach/zig-cache/o/d0bdf9f6e450ee3ba9cbd1a840fb1b36/build /nix/store/z5mqbfapamm3rp7hbr8dcr5jf7kcxkmx-zig-0.10.0-dev.3900+ab4b26d8a/bin/zig /home/desttinghim/Code/mach /home/desttinghim/Code/mach/zig-cache /home/desttinghim/.cache/zig run-example-sysaudio
```

libs/sysaudio/src/main.zig
```diff
@@ -27,8 +27,8 @@ pub const Format = enum {
     F32,
 };
 
-pub const DeviceDescriptor = struct {
-    mode: ?Mode = null,
+pub const DeviceConfig = struct {
```
slimsag, line 30: I propose the following naming:

* `sysaudio.DeviceConfig` -> `sysaudio.Device.Options`
* `sysaudio.DeviceDescriptor` -> `sysaudio.Device.Properties`

This aligns us with WebGPU terminology, which is nice.

examples/sysaudio/main.zig
desttinghim: The sample_rate set here is temporary. Admittedly this is not the best way to handle it. I'll pass the descriptor into the play function instead.

libs/sysaudio/src/soundio.zig
```diff
@@ -264,18 +265,52 @@ pub fn requestDevice(self: Audio, allocator: std.mem.Allocator, config: DeviceDe
...
+            .input => |d| d.layout().channelCount(),
+            .output => |d| d.layout().channelCount(),
+        }),
+        .sample_rate = @intCast(u32, switch (handle) {
+            .input => |d| d.sampleRate(),
+            .output => |d| d.sampleRate(),
+        }),
+    };
+    std.log.info("channels {}", .{descriptor.channels});
+    std.log.info("sample_rate {}\n", .{descriptor.sample_rate});
```
slimsag, line 313: Can you remove these log lines? I don't know why these are here, they shouldn't be.

libs/sysaudio/src/webaudio.zig
```diff
@@ -114,10 +115,12 @@ pub fn requestDevice(audio: Audio, allocator: std.mem.Allocator, config: DeviceD
     }
 
     // TODO(sysaudio): introduce a descriptor type that has non-optional fields.
-    var descriptor = config;
-    descriptor.mode = descriptor.mode orelse .output;
-    descriptor.channels = descriptor.channels orelse default_channel_count;
-    descriptor.sample_rate = descriptor.sample_rate orelse default_sample_rate;
+    var descriptor = DeviceDescriptor{
```
slimsag, line 118: Please remove the TODO above this (which you fixed) :)

examples/sysaudio/main.zig
slimsag: If we have the soundio backend keep track of whether or not we're currently playing, perhaps that resolves that? I suggest adding an internal `playing: bool` field [here](https://github.com/hexops/mach/blob/main/libs/sysaudio/src/soundio.zig#L26-L30) and then [guarding `play` and `pause` here](https://github.com/hexops/mach/blob/f807c85232e6891ac6cde42bd2dff6e58efa5463/libs/sysaudio/src/soundio.zig#L113-L136)

desttinghim: I tweaked your advice and made the field `started` instead. I need to double check that I'm using the API correctly now.

examples/sysaudio/main.zig
desttinghim: I just hadn't thought of it! I've implemented your suggestion

libs/sysaudio/src/soundio.zig
desttinghim: Probably some debugging prints that got forgotten

libs/sysaudio/src/webaudio.zig
desttinghim: Done!

libs/sysaudio/src/soundio.zig
```diff
@@ -21,13 +24,17 @@ else
...
 
     // Internal fields.
     handle: SoundIoStream,
     data_callback: ?DataCallback = null,
     user_data: ?*anyopaque = null,
     planar_buffer: [512000]u8 = undefined,
+    started: bool = false,
+
+    pub const Options = DeviceOptions;
+    pub const Properties = DeviceProperties;
```
slimsag, line 37: we can/should move the actual type definitions here directly I think.

slimsag (approved):

slimsag:
Nice work on this, thanks a bunch for sending this!

PR #526, opened 2022-09-10
fix: make it build with Zig 0.10.0-dev.3567+95573dbee

locriacyber:


alichraghi:
Your zig is old. atm latest version is *`0.10.0-dev.3978+4fd4c733d`*, not *`0.10.0-dev.3567+95573dbee`*

slimsag:
Agreed, please try the latest Zig version mentioned in the README (https://github.com/hexops/mach#supported-zig-version) and let us know if you run into any issues then :)

PR #544, opened 2022-09-16
mach: add `is_app` option to disable gamemode in desktop applications

### LINKED ISSUE #441: mach: add a way for non-game apps to opt-out of linux gamemode support
With #437 just merged, I realize now all Mach apps will enable Linux gamemode if present on the OS. This is not always ideal, as some Mach apps will be desktop apps.

We should make this a `linux_gamemode` option which defaults to `true` similar to how we have `power_preference` for GPUs: https://github.com/hexops/mach/blob/6eaacde4443cb25e17715d15778914ebf33ec066/src/structs.zig#L46-L47

alichraghi:
Closes #441

### DIFF
--- a/src/structs.zig
+++ b/src/structs.zig
@@ -51,6 +51,9 @@ pub const Options = struct {
 
     /// Whether the application has a preference for low power or high performance GPU.
     power_preference: gpu.PowerPreference = .undef,
+
+    /// If set, optimize for regular applications rather than games. e.g. disable Linux gamemode / process priority, prefer low-power GPU (if preference is .undef), etc.
+    is_app: bool = false,
 };
 
 pub const Event = union(enum) {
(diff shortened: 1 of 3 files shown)


silversquirl:
This double negative feels weird, shouldn't it be `is_game: bool = true`?

src/structs.zig
```diff
@@ -51,6 +51,9 @@ pub const Options = struct {
 
     /// Whether the application has a preference for low power or high performance GPU.
     power_preference: gpu.PowerPreference = .undef,
+
+    /// If set, Optimizes for desktop applications. e.g. Disables Linux Gamemode
+    not_game: bool = false,
```
slimsag, line 56: I propose the following:

```
/// If set, optimize for regular applications rather than games. e.g. disable Linux gamemode / process priority, prefer low-power GPU (if preference is .undef), etc.
is_app: bool = false,
```

slimsag (approved):

PR #545, opened 2022-09-16
Set cache directory to ~/.cache/... instead of zig-cache/

locriacyber:


slimsag:
I don't generally like applications writing outside their contained directory. This would also introduce a difference on Windows (use zig-cache) and non-Windows platforms (use `$HOME/.cache`)

`zig-cache` is also in the project directory rather than `$HOME/.cache` for whatever reason, and I expect projects to keep their cache files required for building the project there (or at least in the project directory.)

This would be a different story perhaps if we were talking about very large downloads, but these binaries are ~20-30 MB so I don't think sharing among different projects on your system is a huge benefit.

I'll reject this change as-is, but happy to continue discussing if you feel strongly about this. Feel free to open an issue

locriacyber:
I want to use mach offline, and I have to share cache in some way.

I don't want to use `zig-cache` to cache other things.

PR #548, opened 2022-09-16
mach: better compile error for missing fields in App.

TUSF:
Stage2 Zig ends up coercing structs without any fields into consts. The current error given is not entirely clear to the cause or how to fix the issue for users.


slimsag (approved):
Thank you!

BeauSLM:
Good news, zig's fixed the issue upstream, meaning you don't have to add fields to the struct. I'll make a quick pr so mach doesn't enforce having at least one field in the app.

PR #551, opened 2022-09-17
known-issues: fix wrong command

C-BJ:

### DIFF
--- a/doc/known-issues.md
+++ b/doc/known-issues.md
@@ -23,7 +23,7 @@ Windows does not have symlinks enabled, or Git is not configured to use them. Th
 
 ## Windows: "SSL certificate problem: unable to get local issuer certificate"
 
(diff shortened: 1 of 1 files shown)


doc/known-issues.md
```diff
@@ -23,7 +23,7 @@ Windows does not have symlinks enabled, or Git is not configured to use them. Th
 
 ## Windows: "SSL certificate problem: unable to get local issuer certificate"
 
-This is a curl SSL CA issue, you may want to Google for proper solutions on your system. That said, you can `set CURL_INSECURE=true` and retry to disable SSL verification if you want to workaround the issue.
+This is a curl SSL CA issue, you may want to Google for proper solutions on your system. That said, you can `git config --global http.sslVerify false` and retry to disable SSL verification if you want to workaround the issue.
```
slimsag, line 26: This isn't a correct change.

It may be the case that you have run into *another* issue with a similar wording, which requires the `git config` option? But the problem this was describing needs the existing command unchanged.

PR #555, opened 2022-09-18
gamemode: use zig build

alichraghi:


.gitignore
```diff
@@ -16,3 +16,5 @@ zig-out/
 /build/
 /build-*/
 /docgen_tmp/
+
+/.vscode/
```
slimsag, line 20: Editor directories should be added to your [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c), not here - otherwise we end up having to add every editor's ignored files to all projects which is bad.

slimsag (approved):

PR #560, opened 2022-09-20
basisu: pass CrossTarget to static libs

kubkon:
If we don't do that, we risk generating an invalid archive format for the cross-target.

---


alichraghi:
now, basisu tests can be enabled on non-native platforms
https://github.com/hexops/mach/blob/8026526b997897265ecc503b1b5158081cfaa247/build.zig#L55 https://github.com/hexops/mach/blob/8026526b997897265ecc503b1b5158081cfaa247/build.zig#L64

slimsag (approved):
Thank you thank you!

Issue #532, opened 2022-09-14
mach: support multiple windows on desktop platforms only


slimsag:
I'd like to support multiple windows on desktop platforms (e.g. for the editor to have breakout windows that dock/undock if you drag'n'drop a window onto the main one.)

Of course, multiple windows will not work on most other platforms (at least mobile, consoles, etc.) - WASM would be nice-to-have.

slimsag:
Here's a starting point perhaps?

* Add `Core.createWindow`, it returns a new `*Window` type (which I imagine would largely have identical fields to `Core` today.)
* Add `Window.setOptions`, `Window.setShouldClose`, `Window.getFramebufferSize`, `Window.getSize`, `Window.close`

Then inside your app's `pub fn init` or `pub fn update` you could call this API to create a new window and render to it.

slimsag:
Hey @kchilton - there is a *major* API design change of the `Core.zig` API currently underway, and also our ECS. It's been discussed briefly in the Discord #contributing channel, and I'll have a PR up for it in the next few weeks probably.

After that change we'll be able to support this easily I think.

PR #562, opened 2022-09-20
WIP: multi-window handling on native

praschke:
#532

some fields have been reorganized, as Window, BackingWindow, and WindowOptions share the window-specific interfaces of Core, Platform, and Options. maybe that should be split out on its own.

it might make sense to fully refactor window-specific interfaces out, but i couldn't figure out a nice way to organize it. that may become necessary depending on how Events convey which window an event is for to the user. that's also why i'm not sure about who should be allocating the Window.

### DIFF
--- a/src/Core.zig
+++ b/src/Core.zig
@@ -1,11 +1,11 @@
 const std = @import("std");
 const Allocator = std.mem.Allocator;
 const builtin = @import("builtin");
-const glfw = @import("glfw");
 const gpu = @import("gpu");
 const platform = @import("platform.zig");
 const structs = @import("structs.zig");
 const enums = @import("enums.zig");
+const Window = @import("Window.zig");
 const Timer = @import("Timer.zig");
 
 const Core = @This();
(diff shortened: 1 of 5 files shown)


slimsag:
Sorry for the delay in getting back on this :) I'll be busy the next few weeks, but I'm very excited about this work! This is generally looking headed in the right direction to me 👍 

> some fields have been reorganized, as Window, BackingWindow, and WindowOptions share the window-specific interfaces of Core, Platform, and Options.

These changes in this PR look good to me.

> it might make sense to fully refactor window-specific interfaces out, but i couldn't figure out a nice way to organize it. that may become necessary depending on how Events convey which window an event is for to the user. that's also why i'm not sure about who should be allocating the Window.

Here's been my thinking:

* `Core` owns all windows. By default it has just one window. `Core.initWindow` gives you more windows, which `Core` owns.
  * As far as refactoring code goes, I'm probably in favor of anything you'd advise here.
  * I think I would also be OK with having `core.window` be the _default_ main window, and having some other way to access secondary windows. (`core.windows` slice or something?)
* Just as with GLFW, there is central event handling for all windows. The events themselves should describe which window the event occurred for.

praschke:
this is stale, and i don't have confidence in what changes are suitable. i can try again later if i ever get back to it, or someone else can address it.

PR #572, opened 2022-09-29
freetype: don't expose harfbuzz directly from freetype pkg

mlugg:
This declaration itself was nonsensical, as the file being imported was from the harfbuzz package (instead users should import the harfbuzz package directly). But as well as being nonsensical, this line was occasionally triggering a stage2 bug which we haven't quite tracked down yet which seems to be something to do with conflicting modules importing a file. This issue meant that projects with a specific dependency pattern on freetype and harfbuzz would sometimes randomly get a nonsensical compilation error. This change works around that issue for mach-freetype.


alichraghi (approved):

alichraghi:
CI get fixed after #571

PR #584, opened 2022-10-16
Remove submodules/symlinks in favor of zig-fetch

slimsag:
## Context

Mach today has very few external dependencies. The `examples/` depend on three:

* `zigimg` (image decoding)
* `zmath`
* `assets` (a few e.g. PNG files used for examples)

Otherwise, Mach "owns" all of it's dependencies - shipping the most minimal set of sources on it's own to provide the foundational aspects of graphics, audio, etc.

* `glfw` (window management)
* `basisu` (basis universal supercompressed texture support)
* `soundio` (andrew kelley's libsoundio for audio input/output only)
* `dawn` (webgpu)
* `freetype` (font rendering)

Although we own these sources, we do not commit them to the main repository - as that could pollute the repo history long-term. Instead, we ship these in separate repositories, e.g. https://github.com/hexops/glfw and https://github.com/hexops/freetype - with just the minimal set of C sources needed to build for every OS.

### Problem

Today, we use submodules to ship these dependencies. In some cases, we use a symlink to resolve our own code (in the same repo). This is problematic for two reasons:

1. Symlinks are terrible to work with on Windows.
2. People hate git submodules with a passion.
  a. Note: We handle submodules automagically for users, i.e. if you forgot to clone the submodules `zig build` does it for you.
3. When people see a repository contains submodules, their intuition is to clone the repository recursively. This may normally be a good idea, but in the case of Mach it means you're also cloning the Dawn source code, and it's submodules like DirectXShaderCompiler, and it's submodules, and so on - when in reality you don't need any of this because we default to shipping a binary build of Dawn (unless you specify `-Ddawn-from-source`)
4. Since you need to import Mach's `build.zig`, if using Mach as a library you must use a submodule too.

The third point is the most impactful, causing people to think that Mach is a larger / more heavyweight codebase than it really is in practice. The fourth point is a major annoyance.

### Solution

The ultimate solution will be when Zig gets it's official package manager https://github.com/ziglang/zig/issues/943 (i.e. that will hopefully obsolete all of this work, and make it so that these are non-questions.)

For now, however, we need a better solution.

In [machlibs/synth](https://github.com/machlibs/synth) @desttinghim came up with a rather nice solution, an extension on what we'd done previously (i.e. running the commands to clone git submodules for you if you forgot): first have `build.zig` run the relevant commands to `git clone` repositories (not submodules), then once finished haver it execute _another_ `build.zig` file which is your real one (which `@import`s your dependencies.) This is a good quality of life improvement, and eliminates submodules entirely. @bootradev then ran with this idea and created a library for it, [zig-fetch](https://github.com/bootradev/zig-fetch).

This PR switches us from using submodules and symlinks, to just using zig-fetch. Mach is a bit more of a complex project than most, so some changes to zig-fetch were needed - and so for now we're using a fork (https://github.com/hexops/zig-fetch) which I'd like to upstream eventually.

## How this impacts you

You should be aware that there are no more submodules. This means the following directories are no longer cloned for you by Git itself:

* `mach/libs/basisu/zig-deps/upstream`
* `mach/libs/freetype/zig-deps/upstream`
* `mach/libs/glfw/zig-deps/upstream`
* `mach/libs/sysaudio/zig-deps/upstream` 

This also means you do not need to have submodules in your codebase anymore, either.

### Consumers of Mach core / Mach engine

If you are importing all of Mach (https://github.com/hexops/mach) into your codebase, you should remove the `mach` submodule (`git rm -r libs/mach`) and instead switch to using `zig-fetch`. The project setup is simple: https://gist.github.com/slimsag/106ad310a340781376f16a422aafc816

### Consumers of mach-glfw, mach-freetype

If you are using `mach-glfw` or `mach-freetype` as a submodule in your project, you have two options:

1. Use zig-fetch in your project, which would look like this: https://gist.github.com/slimsag/aea6dfd2c4270f8ba1ed1eb7e56a3bd6
    * Note: You can do this just for Mach libraries to start with, and adopt it incrementally. No need to abandon all your submodules at once.
2. Manually clone these dependencies. For example, [mach-glfw](https://github.com/hexops/mach/blob/f62387c9468641c942c798dd7013e0f8b7e50dcc/libs/glfw/build.zig#L6-L12) only requires you clone `https://github.com/hexops/glfw` into `mach-glfw/zig-deps/upstream`.

## Feedback

I'm looking for feedback before moving forward with this change. Do you like this? dislike this?

In the end, we'll adopt whatever Zig adopts for the official package manager. So this is just a temporary solution until that work begins to surface.

---


slimsag:
Declining this change in favor of https://github.com/hexops/mach/pull/585

PR #590, opened 2022-10-21
mach: implement setCursorMode normal/hidden for wasm

### LINKED ISSUE #577: mach: add WASM support for GLFW-like cursor modes (normal, hidden, disabled)
I merged https://github.com/hexops/mach/commit/78dfa2df6b2d5a38298fe16eaa615ed480433d63 for cursor mode support (e.g. needed for FPS-style games where you need the cursor to stay in the viewport hidden.)

We should have a WASM implementation of this. It may require changing the API to a callback-based one, since IIRC browsers require a permission prompt for this.

alichraghi:
Fixes #577 

### DIFF
--- a/.github/workflows/update_examples.yml
+++ b/.github/workflows/update_examples.yml
@@ -8,6 +8,7 @@ concurrency:
   cancel-in-progress: true
 jobs:
   publish:
+    if: github.repository_owner == 'hexops'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout hexops/mach-examples repo


src/platform/mach.js
```diff
@@ -323,6 +323,16 @@ const mach = {
     }
   },
 
+  machSetCursorMode(cursor_ptr, len) {
+    let mach_name = mach.getString(cursor_ptr, len);
+
+    if (mach_name === 'normal') document.body.style.cursor = 'default';
+    else if (mach_name === 'hidden' || mach_name === 'disabled') document.body.style.cursor = 'none';
```
slimsag, line 330: A "disabled" cursor mode has a different meaning than just hidden. Quoting GLFW docs:

> This will hide the cursor and lock it to the specified window. GLFW will then take care of all the details of cursor re-centering and offset calculation and providing the application with a virtual cursor position. This virtual position is provided normally via both the cursor position callback and through polling.

In specific I think we will need to call `canvas.requestPointerLock()` and `document.exitPointerLock()` here.

We should use these polyfills for better browser support:

```js
canvas.requestPointerLock = canvas.requestPointerLock || canvas.mozRequestPointerLock;
document.exitPointerLock = document.exitPointerLock || document.mozExitPointerLock;
```

We should also add error handling (just `console.error` for now I guess is fine):

```js
document.addEventListener('pointerlockerror', lockError, false);
document.addEventListener('mozpointerlockerror', lockError, false);

function lockError(e) {
  console.error("Pointer lock failed");
}
```

Ideally, we should also add an event handler:

```js
if ("onpointerlockchange" in document) {
  document.addEventListener('pointerlockchange', lockChangeAlert, false);
} else if ("onmozpointerlockchange" in document) {
  document.addEventListener('mozpointerlockchange', lockChangeAlert, false);
}

function lockChangeAlert() {
  if (document.pointerLockElement === canvas ||
  document.mozPointerLockElement === canvas) {
    console.log('The pointer lock status is now locked');
    // Do something useful in response
  } else {
    console.log('The pointer lock status is now unlocked');
    // Do something useful in response
  }
}
```

When this fires, we should send a Mach event to indicate the cursor mode has changed back to "normal".

We may also need to change the mouse event handling code, since the cursor position values are now expected to be different: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API#locked_state

I expect we can align these values with "raw mouse motion" events from GLFW: https://www.glfw.org/docs/latest/input_guide.html#cursor_mode

slimsag (approved):
I will merge since this is still a clear improvement over the current code. Will create a new issue to reflect broken "disabled" cursor support.

PR #611, opened 2022-11-12
mach: add option to set the monitor index on fullscreen

dasimmet:
For multiple monitors (at least on native), it would be nice to be able to select the monitor for fullscreen mode.

### DIFF
--- a/src/platform/native.zig
+++ b/src/platform/native.zig
@@ -319,7 +319,11 @@ pub const Platform = struct {
         if (options.fullscreen) {
             platform.last_position = try platform.window.getPos();
 
-            const monitor = glfw.Monitor.getPrimary().?;
+            const monitorList = try glfw.Monitor.getAll(platform.allocator);
+            defer platform.allocator.free(monitorList);
+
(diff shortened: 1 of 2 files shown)


src/platform/native.zig
```diff
@@ -316,10 +316,16 @@ pub const Platform = struct {
             .double => .fifo,
             .triple => .mailbox,
         };
+
+        platform.last_position = try platform.window.getPos();
+
         if (options.fullscreen) {
-            platform.last_position = try platform.window.getPos();
 
-            const monitor = glfw.Monitor.getPrimary().?;
```
slimsag, line 322: Thanks for sending this change. Happy to merge this.

However, I don't think the default `monitor: u32 = 0` is always going to be equal to `glfw.Monitor.getPrimary()`. I would like for the default to remain the primary monitor.

Can you make `monitor: u32 = 0` instead be `monitor: ?u32 = null` and choose the `getPrimary()` monitor if null?

slimsag (approved):

PR #614, opened 2022-11-14
known-issues: add a solution to the `Windows: File not found` problem

C-BJ:

### DIFF
--- a/doc/known-issues.md
+++ b/doc/known-issues.md
@@ -10,7 +10,7 @@ If you encounter an error like this:
 
 Windows does not have symlinks enabled, or Git is not configured to use them. This is very annoying and [has been reported to Microsoft](https://twitter.com/slimsag/status/1508114938933362688).
 
-**Two solutions:**
+**Three solutions:**
 
 1. (recommended) Build a native Windows binary by cross-compiling from WSL -> Windows:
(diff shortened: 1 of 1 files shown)


slimsag:
It's not clear to me what you are fixing here.

The instructions here already say as the last step: `Re-clone the repository and try again.` - did you run into an issue doing that?

slimsag:
Apologies but this change doesn't make sense to me. Please open an issue if you run into this and I can help / figure out how to update these docs. As it stands, I think this would confuse others.

> didn't download the git submodules, since dawn and DirectXShaderCompiler are removed from the source tree, the problem can be solved by recursively clone

These submodules are not needed. Cloning recursively will not help with cloning them either, as they are not submodules anymore.

C-BJ:
I know that `git clone --recursive` do not clone dawn and DirectXShaderCompiler, `git clone --recursive` just clone the submodules like glfw.

C-BJ:
since dawn and DirectXShaderCompiler are removed from the source tree, `git clone --recursive` will not clone them, so we can use `git clone --recursive` to clone the submodules we need, so the issue in doc should be fixed

C-BJ:
@slimsag

Issue #610, opened 2022-11-09
gpu-dawn: macOS 11 reports 'binaries not available'


belm0:
I tried the canonical build on Big Sur, and it failed on lack of gpu-dawn binaries.  Adding `-Dtarget=x86_64-macos.12` seems to run fine.

Is this expected, or should the build automatically target macos.12 even though it's on 11?

mach-cpu-dawn README seems to imply that the target flag is applicable to any macOS:
> | `-Dtarget=x86_64-macos.12`    | Cross compile to macOS (Intel chipsets)                    |

```
$ zig build run-boids
Cloning into 'sdk-macos-11.3'...
remote: Enumerating objects: 15503, done.
remote: Counting objects: 100% (140/140), done.
remote: Compressing objects: 100% (117/117), done.
remote: Total 15503 (delta 26), reused 23 (delta 23), pack-reused 15363
Receiving objects: 100% (15503/15503), 23.33 MiB | 4.05 MiB/s, done.
Resolving deltas: 100% (2273/2273), done.
Updating files: 100% (12360/12360), done.
error: gpu-dawn binaries for x86_64-macos.11.7.1...11.7.1-none not available.
error: -> open an issue: https://github.com/hexops/mach/issues
error: -> build from source (takes 5-15 minutes):
error:        use -Ddawn-from-source=true or set `Options.from_source = true`
error: -> Did you mean to use -Dtarget=x86_64-macos.12 ?
```

slimsag:
@belm0 Very interesting that you're able to run an application targeting macOS 12 on macOS 11! I didn't think that would work (I mean, it's all dynamic linking to system frameworks, so there's always a chance it works.. I guess we just got lucky.) Thanks for mentioning this.

I need to update our infrastructure to better support all 3 latest macOS versions that Zig supports. I'll do this soon. Off the top of my head, this includes:

* [ ] Testing we can build for all 3 latest versions
* [ ] Creating a macOS 13 SDK
* [ ] Updating gpu-dawn binary builds to produce binaries for all 3 latest versions.

In the meantime, I am happy to merge PRs that may improve the situation-such as by relaxing our version constraints to allow macOS 13 to use the macOS 12 SDK.

PR #619, opened 2022-11-18
sysaudio: add ability to provide system_sdk options

mifopen:
Do not really know if it's ok or not, but for now, it is the only way to build Mach under macOS 13 without preparing actual SDK (which is a hard task for me). With this change, one can pass overridden `sdk_list` to glfw and sysaudio libs.
Btw on my laptop it is enough to copy `Sdk` from 12th version and just change semver to 13 to successfully compile Mach.
Can be used as a temporary workaround for https://github.com/hexops/mach/issues/610

### DIFF
--- a/libs/sysaudio/sdk.zig
+++ b/libs/sysaudio/sdk.zig
@@ -17,6 +17,9 @@ pub fn Sdk(comptime deps: anytype) type {
 
         pub const Options = struct {
             install_libs: bool = false,
+
+            /// System SDK options.
+            system_sdk: deps.system_sdk.Options = .{},
         };
 
         pub fn testStep(b: *std.build.Builder, mode: std.builtin.Mode, target: std.zig.CrossTarget) *std.build.RunStep {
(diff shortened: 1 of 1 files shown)


mifopen:
For example
```zig
    const system_sdk_options = .{
        .sdk_list = &.{
            .{
                .name = "sdk-macos-12.0",
                .git_addr = "https://github.com/hexops/sdk-macos-12.0",
                .git_revision = "14613b4917c7059dad8f3789f55bb13a2548f83d",
                .cpu_arch = &.{ .aarch64, .x86_64 },
                .os_tag = .macos,
                .os_version = .{
                    .semver = .{
                        .min = .{ .major = 13, .minor = 0 },
                        .max = .{ .major = 13, .minor = std.math.maxInt(u32) },
                    },
                },
            },
        },
    };
    try app.link(.{
        .glfw_options = .{ .system_sdk = system_sdk_options },
        .sysaudio_options = .{ .system_sdk = system_sdk_options },
    });
```

slimsag (approved):

slimsag:
Thank you very much for sending this change. It looks good, and it is useful to pass through these options regardless of the issue you're working around.

PR #621, opened 2022-11-22
freetype: add option to build using separate zlib

david-vanderson:
I ran into this when trying to use both mach-freetype and zig-libcurl (https://github.com/mattnite/zig-libcurl).  First I got undefined symbols like:

```
ld.lld: error: undefined symbol: inflate
>>> referenced by content_encoding.c:196 (libs/zig-libcurl/curl/lib/content_encoding.c:196)
>>>               /home/dvanderson/gui/zig-cache/o/e7a848841ad0f6ef64dd93429b6b18b5/content_encoding.o:(inflate_stream) in archive /home/dvanderson/gui/zig-cache/o/95294db893708419c83e8414f31f1147/libcurl.a
```
So I added zig-zlib (https://github.com/mattnite/zig-zlib).  Then I got duplicate symbols like:
```
ld.lld: error: duplicate symbol: inflateResetKeep
>>> defined at inflate.c:123 (libs/mach/libs/freetype/upstream/freetype/src/gzip/inflate.c:123)
>>>            /home/dvanderson/gui/zig-cache/o/6a165bd051f981ac91cbb4f2624cd006/ftgzip.o:(inflateResetKeep) in archive /home/dvanderson/gui/zig-cache/o/1805910d2fe891e313ab5e3835f3c7d4/libfreetype.a
>>> defined at inflate.c:121 (libs/zig-zlib/zlib/inflate.c:121)
>>>            /home/dvanderson/gui/zig-cache/o/9b6e8398cc498c39e731f1280b430117/inflate.o:(.text+0x0) in archive /home/dvanderson/gui/zig-cache/o/342308fbe7d19c8c086056be2b44f141/libz.a
```
My best guess so far is that freetype normally builds its own copy of zlib in a way that doesn't expose everything?  But enough to conflict with a separate zlib?

This exposes the option FT_CONFIG_OPTION_SYSTEM_ZLIB already present in freetype.  I think this makes sense but let me know if I'm going about this the wrong way.

Thanks!

### DIFF
--- a/libs/freetype/README.md
+++ b/libs/freetype/README.md
@@ -30,6 +30,9 @@ pub fn build(b: *Builder) void {
     ...
     exe.addPackage(freetype.pkg);
     freetype.link(b, exe, .{});
+
+    // use this option if you are including zlib separately
+    //freetype.link(b, exe, .{ .freetype = .{ .use_system_zlib = true } });
 }
 ```
 
--- a/libs/freetype/build.zig
+++ b/libs/freetype/build.zig
@@ -30,6 +30,7 @@ pub const FreetypeOptions = struct {
     config_path: ?[]const u8 = null,
     install_libs: bool = false,
     brotli: bool = false,
+    use_system_zlib: bool = false,
 };
 
 pub const HarfbuzzOptions = struct {
@@ -131,6 +132,9 @@ pub fn buildFreetype(b: *Builder, mode: std.builtin.Mode, target: std.zig.CrossT
 
     const lib = b.addStaticLibrary("freetype", null);
     lib.defineCMacro("FT2_BUILD_LIBRARY", "1");
+    if (options.use_system_zlib) {
+        lib.defineCMacro("FT_CONFIG_OPTION_SYSTEM_ZLIB", "1");
+    }
     lib.setBuildMode(mode);
     lib.setTarget(target);
     lib.linkLibC();


alichraghi (approved):

slimsag (approved):
Seems reasonable to me.

If you do end up wanting to debug this further the other way around, let me know, we can help with that. It does look like freetype's version of zlib includes the `inflate` function:

https://github.com/hexops/freetype/blob/d2eb35016c56cc3e2ef4076e8d61e3fd035071ab/freetype/src/gzip/zlib.h#L402

Anyway if this works for you, LGTM.

david-vanderson:
I'll spend some more time trying to figure out why I got those linker errors, because you're right that freetype's zlib does include them.

Thanks very much!

david-vanderson:
https://gitlab.freedesktop.org/freetype/freetype/-/issues/1146

After digging into it a bit, it seems like this is a bug in freetype when compiling a static library using the bundled zlib (which seems like not very many people have done),

The underlying problem is that zlib is implemented across multiple source files, which means some of it's functions must be usable across compilation units, which can cause duplicate symbol errors.  zlib uses this Z_PREFIX macro thing to rename all of its functions to work around this.

Freetype compiles them all in a single source file, and tries to make all the zlib functions "static" so they are not visible outside that compilation unit.  It does this by overriding the ZEXTERN macro (to make it "static") that's applied to zlib function declarations.  But that is not enough.  If I add it also to the function definitions, then the duplicate symbols go away.

I will try suggesting that on the freetype issue (once I'm able to comment there).

I still don't understand why only some of the symbols caused duplicates.

david-vanderson:
@slimsag You inspired me to push for it - I was able to work with the freetype maintainers to fix this issue!  When we update our freetype sources, compiling it statically (including bundled zlib) should no longer export (leak) zlib symbols.

Short story:
- freetype patches zlib sources to try and make all functions static
- zlib added new functions at some point
- the freetype patches needed to be updated to make it work, and that's now done

PR #630, opened 2022-11-27
mach: gfx: Add VertexWriter

kdchambers:

### DIFF
--- a/src/gfx/util.zig
+++ b/src/gfx/util.zig
@@ -1,0 +1,191 @@
+const std = @import("std");
+
+/// Vertex writer manages the placement of vertices by tracking which are unique. If a duplicate vertex is added
+/// with `put`, only it's index will be written to the index buffer.
+/// `IndexType` should match the integer type used for the index buffer
+fn VertexWriter(comptime VertexType: type, comptime IndexType: type) type {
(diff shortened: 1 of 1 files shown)


src/gfx/util.zig
```diff
@@ -0,0 +1,191 @@
...
+
+        pub fn vertexBuffer(self: @This()) []Vertex {
+            return self.vertices[0..self.next_packed_index];
+        }
+    };
+}
+
+const Vec4 = [4]f32;
+const Vec3 = [3]f32;
+const Vec2 = [2]f32;
```
alichraghi, line 103: wouldn't be better to use `@Vector`?

kdchambers: Could be, although this is only a test. I guess it would come down to which is easier to read / a more common use-case. I don't have any issues changing to @Vector.

slimsag (approved):

src/gfx/util.zig
Luexa: An issue to keep in mind with `@Vector` is that `@Vector(3, T)` has the same alignment/padding requirements as `@Vector(4, T)`, so it is not necessarily ideal for vertex buffers.

PR #632, opened 2022-11-27
WIP: mach: add total time tracker

SergeiSomin:
- [+] By selecting this checkbox, I agree to license my contributions to this project under the license(s) described in the LICENSE file, and I have the right to do so or have received permission to do so by an employer or client I am producing work for whom has this right.

### DIFF
--- a/src/Core.zig
+++ b/src/Core.zig
@@ -21,6 +21,11 @@ options: structs.Options,
 /// (360.0 * core.delta_time)
 delta_time: f32 = 0,
 delta_time_ns: u64 = 0,
+
+/// The amount of time (in seconds) that has passed since the start of the application
+total_time: f32 = 0,
+total_time_ns: u64 = 0,
+
 timer: Timer,
 
 device: *gpu.Device,
--- a/src/platform/native.zig
+++ b/src/platform/native.zig
@@ -572,32 +572,28 @@ pub const Platform = struct {
 };
 
 pub const BackingTimer = struct {
-
-	start_time: f64 = undefined,
+    start_time: f64 = undefined,
     timer: std.time.Timer = undefined,
 
-	pub fn start() std.time.Timer.Error!BackingTimer {
-		return BackingTimer{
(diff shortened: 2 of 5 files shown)


src/Core.zig
```diff
@@ -21,6 +21,11 @@ options: structs.Options,
 /// (360.0 * core.delta_time)
 delta_time: f32 = 0,
 delta_time_ns: u64 = 0,
+
+/// The amount of time (in seconds) that has passed since the start of the application
+total_time: f32 = 0,
+total_time_ns: u64 = 0,
```
slimsag, line 27: Instead of exposing these two as fields here, I suggest users just call the function themselves for now. We don't know in practice how many people will use this API, and updating these fields every frame isn't free.

src/platform/native.zig
```diff
@@ -571,7 +571,30 @@ pub const Platform = struct {
...
+    pub fn start() std.time.Timer.Error!BackingTimer {
+        return BackingTimer{ .time = try std.time.Timer.start(), .start_time = std.time.Instant.now() };
+    }
+
+    pub fn read(self: *BackingTimer) u64 {
+        return self.timer.read();
+    }
+
+    pub fn readTotal(self: *BackingTimer) u64 {
+        return std.time.Instant.now() - self.start_time;
```
slimsag, line 587: `std.time.Instant.now()` is not guaranteed to return monotonic time values. That is, it may return a time that is *before* `self.start_time` if a user e.g. sets their system clock back in time. In which case, this would overflow.

Instead of storing `start_time` and doing this, I suggest instead we just keep two `std.time.Timer` fields around (as they are guaranteed to be monotonic):

```diff
-start_time: f64 = undefined,
+start_timer: std.time.Timer = undefined,
 timer: std.time.Timer = undefined,
```

Then this `readTotal` function should just be `return self.start_time.read()`.

slimsag:
https://github.com/hexops/mach/issues/824

PR #634, opened 2022-12-10
use try to handle build.Pkg oom errors

dasimmet:
Using newer versions of zig (tested with `0.11.0-dev.664+c49e4d534`)  `ArrayList.toOwnedSlice` returns an error union:

```zig
fn toOwnedSlice(self: *Self) Allocator.Error!Slice
```

which causes a build error on the dependencies ArrayList, which needs to be handled. It would break compatibilty with 0.10.0 though.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -220,7 +220,7 @@ pub const App = struct {
         const app_pkg = Pkg{
             .name = "app",
             .source = .{ .path = options.src },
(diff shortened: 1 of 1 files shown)


slimsag:
Sorry, I merged another PR which fixes this by someone else: https://github.com/hexops/mach/pull/635

PR #637, opened 2022-12-13
all: use null and undefined in enums

### LINKED ISSUE #626: gpu: use "null" and "undefined" enums
It's now possible to use these as enum field names, so we can replace `nul` and `undef`

BratishkaErik:
Closes #626
Signed-off-by: Eric Joldasov <bratishkaerik@getgoogleoff.me>

`.dimension_undef`, `limit_u32_undef` and similar was not touched.


BratishkaErik:
Guess I should split commits

BratishkaErik:
Did I split them right?

slimsag (approved):
Thanks, this looks great!

Would you also be interested in sending a PR to update the examples? https://github.com/hexops/mach-examples

Issue #564, opened 2022-09-21
glfw: tracking upstream issues


alichraghi:
following changes must be applied after 3.4 release
- glfw/glfw#2186
    https://github.com/hexops/mach/blob/6ce4110baceaebb0d6d895ffbcf729e1a5d4c5b8/libs/glfw/src/Monitor.zig#L319-L321
    ```diff
    -- /// TODO: Is the documentation obsolete? On wayland the error returned is FeatureUnavailable 
    ```

PR #639, opened 2022-12-14
WIP: glfw: Calling pollEvents can reach unreachable code.

dumheter:
Calling pollEvents on windows will cause Error.InvalidValue when scancode 256 is hit on keyboard, which is the volume up key. It would reach unreachable code.

### DIFF
--- a/libs/glfw/src/main.zig
+++ b/libs/glfw/src/main.zig
@@ -372,12 +372,13 @@ pub fn platformSupported(platform: PlatformType) bool {
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: events, glfw.waitEvents, glfw.waitEventsTimeout
-pub inline fn pollEvents() error{PlatformError}!void {
+pub inline fn pollEvents() error{PlatformError,InvalidValue}!void {
     internal_debug.assertInitialized();
     c.glfwPollEvents();
     getError() catch |err| return switch (err) {
         Error.NotInitialized => unreachable,
         Error.PlatformError => |e| e,
+        Error.InvalidValue => |e| e,
         else => unreachable,
     };
 }


slimsag:
Nice find!

Are you interested in sending an issue upstream to GLFW? It looks like this is not documented behavior: https://github.com/glfw/glfw/blob/dd8a678a66f1967372e5a5e3deac41ebf65ee127/include/GLFW/glfw3.h#L4414

If not I can do so and will then merge this with a comment added to our code:

```
// TODO(glfw): follow up on whatever the outcome of <link> is
```

alichraghi:
https://github.com/hexops/mach/issues/564

dumheter:
I tried to make a repro directly in the GLFW example project, but couldn't, tried both from master and the one you use.

I tried to make a minimal example from my code, but then its working fine as well.

I tried to surround the pollEvents with glfwGetError calls in my repro case, but it really is the pollEvents that is triggering the InvalidValue.

I don't really know a good next step. Would be nice to have the debugger follow into the c code, but I can't do that with RemedyBG on windows.

alichraghi:
why are you trying to repro it?

dumheter:
With glfw having 400 issues open, and 100 pr's open, I feel like it would be a noop to report without more information than "it happened on my machine". But maybe that is fine?

slimsag:
@dumheter with which codebase are you able to reproduce this on Windows? I can take a look soon. I fear that perhaps `pollEvents` is _not_ the cause here but rather the error is being generated elsewhere and not being handled properly there (perhaps in another thread even), and `pollEvents` is just the first one to check if a GLFW error is present

dumheter:
I think so too. I'll come back if I figure it out. (Dont want to waste your time with my codebase).

slimsag:
Closing for now, feel free to reopen or create an issue if you still encounter any problems here :) happy to help however I can

PR #642, opened 2022-12-17
ecs: store column values as independent arrays

Byteron:
I made minimal changes to make the archetype have a block per component type, rather than one large block.

there is 2 relevant commits to this. (1 is format 4 is merge due to double format)

first I extracted fetching the column values into separate functions getColumnValues and getRawColumnValues.

The whole fetching process is done multiple times in other functions, so I figured it makes sense to extract that to a separate function, so I have to make the block change once, instead of multiple times.

second commit is the actual block -> blocks change.

Note that I had to tweak the test slightly: 
for easier column access, the columns are sorted by type_id:
```
for (storage.columns) |column, i| {
  // fetch the block for that column
  const column_block = storage.blocks[i];
}
```


slimsag:
Sorry it took me so long to get to reviewing this :) I wanted to make sure I digested it fully and distractions got the better of me. The addition of `getColumnValues` and `getRawColumnValues` looks great 👍 

## Failing test

I noticed one of the tests was failing (undefined behavior):

```
slimsag@hexops-ci ecs % zig build test
Test [2/5] test.example... The following command terminated unexpectedly:
cd /Users/slimsag/Desktop/hexops/mach-examples/libs/mach/libs/ecs && /Users/slimsag/Desktop/hexops/mach-examples/libs/mach/libs/ecs/zig-cache/o/f2f01dd18a3f30979df199c759f3a713/ecs-tests /Users/slimsag/zig-macos-aarch64-0.11.0-dev.945+0c30e006c/zig 
error: the following build command failed with exit code 1:
/Users/slimsag/Desktop/hexops/mach-examples/libs/mach/libs/ecs/zig-cache/o/7f3d7435806f6e364f470bcd2896457f/build /Users/slimsag/zig-macos-aarch64-0.11.0-dev.945+0c30e006c/zig /Users/slimsag/Desktop/hexops/mach-examples/libs/mach/libs/ecs /Users/slimsag/Desktop/hexops/mach-examples/libs/mach/libs/ecs/zig-cache /Users/slimsag/.cache/zig test
```

The problem seems to occur inside of `setRow`, and after some digging I found out why: column order changes over time (since columns are sorted), but block order is seemingly never updated. `setRow` then expects that blocks can be indexed by column indices - which then leads to completely wrong blocks of memory being associated with different columns.

The more I look at it, I think it's strange that we would keep an `[]Column` and a separate `[][]u8` (array of per-column row values) separately and try to keep those indexed in the same way. We should just keep a `values: []u8` field inside of `Column`, then we have a guaranteed ordering we can't screw up.

I haven't yet dove into fixing this, but I can later unless you prefer giving it a shot.

## Other things I changed

### Removed reliance on `ArchetypeStorage.allocator`

In a few places `gpa: Allocator` arguments were removed from `ArchetypeStorage` methods in this change, in favor of relying on `storage.allocator`.

I would like `ArchetypeStorage` to become an unmanaged data structure (i.e. it doesn't hold a pointer to its allocator so as to not waste memory if we have many small ECS tables.)

I undid this change and removed `allocator` entirely from `ArchetypeStorage`.

### Zero size ColumnType check

In `ArchetypeStorage.set` the condition `if (@sizeOf(ColumnType) == 0) return;` was removed, I added it back. This is important for making `set` no-op in the case the value we're setting is zero size (i.e. void, the component is merely a tag set on the entity)

### Reliance on component names

In a few places (namely `ArchetypeStorage.set`, `ArchetypeStorage.get`), the component `name` was removed in favor of relying on a type ID. I understand why this was done, and I also want to say clearly that I agree the current `name` reliance is not good: it's expensive, and we need a better name -> integer mapping approach.

However, this is a step backwards in my eyes. Before this change we only actually stored type ID information purely for the sake of Zig type-checking in debug mode. A theoretical GUI editor, or C ABI, could supply their own made-up type IDs (or none at all) and our ECS would be none the wiser.

After this change, we increase reliance on type ID which is not acceptable for me because it is prohibitive towards developing a GUI editor, C ABI, etc. I have another idea for how we can fix this with similar performance without relying on typeID, so I've undone this change.

### Minor style things

I fixed a few other minor style things (e.g. if statements with only one line in their code block can be collapsed into a single line)

slimsag (changes requested):

slimsag (approved):

PR #644, opened 2022-12-22
mach: fix compiler error about missing field_type

citruslee:
I made a change in order to fix the compilation (at least on m1 macs) with latest zig (0.11.0-dev.926+266e2e9a3). 


slimsag (approved):
Awesome, thanks for this!

PR #645, opened 2022-12-24
fix compiler errors regarding zig changes

cykrr:
I changed some stuff that didn't allowed me to build mach-examples (`zig build`) on `0.11.0-dev.923+a52dcdd3c`, very similar to previous PR. I was not sure about removing  `.layout = .Auto` as i have no idea about what it does, but it does fix the build.


InKryption:
Removing `.layout = .Auto` is indeed the correct fix. The `layout` is from a time long past when `extern enum` was a thing; it's since been removed because it was too complex, and then it was just a matter of nobody changing the struct in std.builtin and all of the relevant compiler code to reflect the change until recently.

slimsag (approved):

PR #657, opened 2022-12-31
gpu: Fix a few compile errors for the sample_utils file in examples

tlein:
Tried running `zig build` in the `libs/gpu` director (using the version of zig that the CI scripts are setup to use: `0.11.0-dev.1000+94780f7cd`) and got a few errors from the `examples/sample_utils.zig` file. Easy enough to fix, though I don't know yet enough about the project to say if this is necessary. `zig build` and `zig build run-example` work in the `gpu` directory after this fix though.


slimsag:
Did you mean to close this? The changes here look valid/correct, I'd love to merge this if you reopen and are OK with clicking the checkbox :)

slimsag:
in any case thanks for looking at Mach / giving it a try!

PR #661, opened 2023-01-01
all: build: fix sdkPath for relative @src.file / fix autocompletion with ZLS / IDEs

Luexa:
Prior to this commit, the build system heavily assumed that the result of `@src.file` would always be absolute, but this is no longer guaranteed, likely due to there being no such thing as an "absolute path" in WASI.

It appears that for normal invocations of `zig build`, it is safe to assume that `@src.file` is absolute. However, when ZLS uses a custom `build_runner.zig` to collect build configuration, `@src.file` is actually relative to the current working directory, at least on my system. For a while, this led to ZLS completions breaking entirely, but presently it actually causes ZLS to crash!

The solution is not as simple as using relative `sdkPath` results as-is, because the build system may attempt to resolve these paths relative to build root, when the paths are actually relative to the current working directory.

This leads to a sticky situation: the current working directory is a runtime concept, but `@src.file` is resolved at compile time. However, it appears that the build runner does not change current working directory in between compilation and execution, so it is probably safe to calculate `sdkPath` using runtime current working directory.

Still, this requires major changes with how `sdkPath` works, since runtime computation and allocations are required. So pretty much anything that relied on `sdkPath` being comptime-known has been refactored in this commit.

The most severe result of this is that, for example, `gpu.pkg` can no longer be a comptime-known constant: it has to be a runtime function that takes a `*Builder` and returns a `Pkg`.

This commit deals with usages of `*.pkg` and `sdkPath` within Mach itself, but projects that depend on Mach such as `mach-examples` will almost certainly require changes as well.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -27,11 +27,27 @@ const CrossTarget = std.zig.CrossTarget;
 const Builder = std.build.Builder;
 const Pkg = std.build.Pkg;
 
-pub const pkg = Pkg{
-    .name = "mach",
-    .source = .{ .path = sdkPath("/src/main.zig") },
-    .dependencies = &.{ gpu.pkg, ecs.pkg, sysaudio.pkg, earcut.pkg },
-};
+var cached_pkg: ?Pkg = null;
+
+pub fn pkg(b: *Builder) Pkg {
+    if (cached_pkg == null) {
+        const dependencies = b.allocator.create([4]Pkg) catch unreachable;
+        dependencies.* = .{
+            gpu.pkg(b),
+            ecs.pkg(b),
+            sysaudio.pkg(b),
+            earcut.pkg(b),
+        };
+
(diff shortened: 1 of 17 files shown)


slimsag:
Thank you for taking the time to investigate and fix this. It is unfortunate this requires some changes in user code, but I think having auto-completion work in your IDE is well worth it.

I've updated the READMEs to make it easy for anyone to update, luckily it looks like a rather small change for general usage: https://github.com/hexops/mach/pull/661/commits/88b1106953e632735438b352048a7a9e65f3e866

I also confirmed all mach-examples build as-is (no changes needed.)

slimsag (approved):

Luexa:
Oh good to know! I guess because mach-examples uses the "app" API rather
than directly registering the pkgs.

PR #662, opened 2023-01-02
glfw: complete overhaul of error handling (remove Zig errors)

slimsag:
## The problem

When I wrote mach-glfw (one of my earliest Zig projects, in fact) - the hope was that we could map GLFW errors to nice Zig errors. This is not a new approach to me, in fact years ago I was one of the original authors of the best GLFW bindings in Go where we used a similar approach, and encouraged GLFW to more clearly document errors it can return etc.

The problem is that GLFW still today is very much not designed/intended to be used in a way where errors are handled per-function-call. The general expectation is that you will check errors at important points, or via a callback. 

* GLFW documentation of what errors specific functions can generate is not always 100% correct, and often errors can be rather vague (e.g. `GLFW_PLATFORM_ERROR`)
* Even when a function is documented as possibly producing a specific type of error, it is not clear exactly how it should be handled. For example, `glfwSwapBuffers` _can_ generate a platform error but it is not exactly specified what one should do in that circumstance.
* Often times, exiting the application or terminating on error is _not_ the right call to make. For example, there are [many simple functions](https://github.com/hexops/mach-glfw/tree/a9e9ac095509fdf6e7002b4c49e2d8c736dcfe5c#a-warning-about-error-handling) not supported with Wayland, or even on specific X11 compositors, which will generate an error. In most of the usage scenarios with these APIs, it is best for the user if you merely log/print the error and move on continuing to attempt to run. **Your GLFW application will not support the long-tail of X11 compositors and Wayland Linux distributions otherwise.**

Additionally, since messages cannot be attached to errors in Zig (whether you like this or not) this is extremely important in GLFW error handling, as it often communicates a vague error code followed by a very specific error message. If your application is not printing the error message today, then your error handling is most likely _very bad for the user and effectively no better than your app crashing_.

All of these points combined add up to one thing: our GLFW bindings prior to this change make it easy to shoot yourself (and your users) in the foot by writing `try` in front of a GLFW function call - because we try to map GLFW errors to per-function Zig errors.

## The solution

All per-function error handling has been removed. Instead, you should now handle errors similarly to how one would in GLFW's C API.

## "I don't care, how do I update my code?"

If you really do not care about proper error handling and just want your code to compile again, you should simply remove `try` in front of most of your GLFW function calls. By doing this, your application will actually behave better for most of your users than having had `try` in front of function calls before.

## "How do I handle errors properly?"

After making a GLFW function call, you should check errors explicitly using one of:

* `glfw.getErrorString()` - returns an error message or `null` if none present
* `glfw.getError()` - returns a Zig error or void if none is present
* `glfw.mustGetError()` - returns a Zig error or panics if none is present
* `glfw.setErrorCallback` - if you just want to log your errors (I do NOT suggest terminating your application if this callback is invoked.)

Here are some specific examples:

### Initialization

Previously you had:

```zig
try glfw.init(.{});
```

Proper error handling now looks something like:

```zig
if (!glfw.init(.{})) {
    std.log.err("failed to initialize GLFW: {?s}", .{glfw.getErrorString()});
    std.process.exit(1);
}
```

### Creating a window

`glfw.Window.create` returns `null` in the event of an error, you can handle this by presenting the `glfw.getErrorString()` value to the user.

### Other GLFW API calls

In general, you should be explicit about handling errors. Ask yourself: if this code fails, does my application _need_ to exit, or can I continue best-effort? If your application needs to exit because the API call was critical, then add an error check to present the problem:

```zig
glfw.foobar();
if (glfw.getErrorString()) |msg| {
    std.log.err("glfw.foobar: {s}", .{msg});
    std.process.exit(1);
}
```

Keep in mind that GLFW errors are stored per-thread. If you call a function and it produces an error and you do not retrieve it, then the next person who checks for an error will find it. This means that you may find it useful to be a good citizen by "clearing" any stored errors that you do not care about when finished calling GLFW functions, e.g. using defer:

```
defer getError() catch {}; // clear any errors we generated
```

## "I don't like this"

I don't either, but I have only two options:

1. I can sell you a `try` pipe dream in which GLFW errors can be treated the same as Zig errors, and your application will have poor error handling and your users (especially those using Linux) will suffer.
2. I can remove footguns and provide the best guidance possible to ensure you handle errors in a way that makes your application run on the widest array of systems without fault.

## "I still object!"

Realistically, this doesn't affect Mach itself. I am confident that I know how to avoid all these footguns in Mach when we make use of GLFW and expose an API where it doesn't affect Mach engine or Mach core users.

I am doing this purely for users of mach-glfw. If the users of mach-glfw don't like this change, I may not merge it - but I think it'd be a pretty glaring violation of the `zig zen`.

### DIFF
--- a/libs/glfw/README.md
+++ b/libs/glfw/README.md
@@ -102,17 +102,29 @@ Now in your code you may import and use GLFW:
 ```zig
 const glfw = @import("glfw");
 
+/// Default GLFW error handling callback
+fn errorCallback(error_code: glfw.Error, description: [:0]const u8) void {
+    std.log.err("glfw: {}: {s}\n", .{ error_code, description });
+}
+
 pub fn main() !void {
-    try glfw.init(.{});
+    glfw.setErrorCallback(errorCallback);
+    if (!glfw.init(.{})) {
+        std.log.err("failed to initialize GLFW: {?s}", .{glfw.getErrorString()});
+        std.process.exit(1);
+    }
     defer glfw.terminate();
 
     // Create our window
-    const window = try glfw.Window.create(640, 480, "Hello, mach-glfw!", null, null, .{});
+    const window = glfw.Window.create(640, 480, "Hello, mach-glfw!", null, null, .{}) orelse {
+        std.log.err("failed to create GLFW window: {?s}", .{glfw.getErrorString()});
+        std.process.exit(1);
+    };
     defer window.destroy();
 
     // Wait for the user to close the window.
     while (!window.shouldClose()) {
-        try glfw.pollEvents();
+        glfw.pollEvents();
     }
 }
 ```
--- a/libs/glfw/src/Cursor.zig
+++ b/libs/glfw/src/Cursor.zig
@@ -106,23 +106,19 @@ pub const Shape = enum(i32) {
 /// @param[in] yhot The desired y-coordinate, in pixels, of the cursor hotspot.
 /// @return The handle of the created cursor.
 ///
+/// Possible errors include glfw.Error.PlatformError and glfw.Error.InvalidValue
+/// null is returned in the event of an error.
+///
 /// @pointer_lifetime The specified image data is copied before this function returns.
 ///
 /// @thread_safety This function must only be called from the main thread.
 ///
 /// see also: cursor_object, glfw.Cursor.destroy, glfw.Cursor.createStandard
-pub inline fn create(image: Image, xhot: i32, yhot: i32) error{ PlatformError, InvalidValue }!Cursor {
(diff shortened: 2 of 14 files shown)


InKryption:
Sad to see it go, but it does make sense.

star-tek-mb:
Just an idea. Maybe we could use zgl (zig opengl bindings) approach? Either use assertion (default in debug mode), logging, or do nothing. I like this approach, because it helped me to find an annoying bug in my code (i got nice traceback and fixed issue).

slimsag:
Feedback on this has been overwhelmingly positive, so I will merge.

@star-tek-mb great point, I have update the README with an example on how to use `glfw.setErrorCallback` which effectively provides the same thing (and lets the user choose the behavior.)

PR #663, opened 2023-01-03
gpu: fix example build

apppppppple:
Decided to try since I was bored, turned out to be pretty easy. Don't know if it's even worth merging but ¯\\\_(ツ)\_/¯

Fixes: #659


slimsag (approved):
Many thanks!

PR #664, opened 2023-01-04
mach: fundamental changes

### LINKED ISSUE #528: mach: can resize window to size zero on windows, crashes as a result
also might happen during minimization of window? may need to catch illegal size swap chains and go no-op if that is the case:

```
error: gpu: validation error: Swap Chain size (width: 0, height: 0) is empty.
 - While validating [SwapChainDescriptor "main swap chain"]
 - While calling [Device].CreateSwapChain([SwapChainDescriptor "main swap chain"]).
```

### LINKED ISSUE #623: mach: setOptions API needs a better "update" approach
As it stands today, usage like:

```
core.setOptions(.{.vsync = .double});
```

Is almost always incorrect, because it means you are resetting all other options to their defaults. The correct version would be:

```
var options = core.options;
options.vsync = .double;
core.setOptions(options);
```

We should:

1. Make `core.setOption(.vsync, .double)` a thing, inform users that `setOptions` (plural) is probably not what they want.
2. For the C ABI, document this quirk or something, suggest languages wrapping our C API to implement a `setOption` API of their own or something.

alichraghi:
Fixes #528 and Fixes #623

Key changes:

- `Core` no longer depends on your `App` anymore. Instead, Mach calls your app init/update/deinit functions and your app should manage a `*mach.Core` instance.
- `setOptions` has been replaced with individual functions (`setTitle`, `setSize`, etc)
- ..and more

## Update guide

1. Remove the `core: *mach.Core` parameter from your `App`'s `init`, `deinit` and `update` functions.
2. Add a new `core: mach.Core` field to your `App` struct, initialize it using `mach.Core.init()` and call `mach.Core.deinit()` inside your `App`'s `deinit` callback.
3. Your `App`'s `update` function should now return `!bool` instead of `!void`, if the function returns `true` then Mach will exit.
4. The `resize()` callback no longer exists, instead it has been replaced with a new `framebuffer_resize` event.

Other notable mentions:

1. `core.close()` has been removed; return `true` in `update()` to exit your app instead
2. `core.delta_time` has been removed, you can create your own cross-platform timer with `mach.Timer.start()` and call `my_timer.lap()` in every `update()` call (frame) to get the same result.
3. The following have been renamed:
    ```zig
    Core.device -> Core.device()
    Core.swap_chain_format -> Core.descriptor().format
    Core.current_desc -> Core.descriptor()
    Core.swap_chain.? -> Core.swapChain()
    Core.getWindowSize() -> Core.size()
    ```

Also see the `mach-examples` PR changes for examples, let us know if you run into any issues upgrading we're happy to help.


slimsag:
Many thanks for sending this, can you update to account for proper error handling described in #662 ?

slimsag:
sorry I still haven't gotten to review yet, hoping tomorrow will be the day :)

shaderexp/main.zig
```diff
@@ -75,21 +80,24 @@ pub fn init(app: *App, core: *mach.Core) !void {
...
-pub fn deinit(app: *App, core: *mach.Core) void {
+pub fn deinit(app: *App) void {
     app.fragment_shader_file.close();
-    core.allocator.free(app.fragment_shader_code);
+    allocator.free(app.fragment_shader_code);
 
     app.uniform_buffer.release();
     app.bind_group.release();
+
+    app.core.deinit();
```
slimsag, line 90: I think we should standardize on having this be a `defer app.core.deinit()` in all examples

src/platform/wasm/Core.zig
```diff
@@ -0,0 +1,297 @@
...
+}
+
+pub fn setBorder(self: *Core, value: bool) void {
+    _ = self;
+    _ = value;
+}
+
+pub fn border(self: *Core) bool {
+    _ = self;
+    return undefined;
```
slimsag, line 200: should return `false` not an undefined value

src/platform/wasm/Core.zig
```diff
@@ -0,0 +1,297 @@
...
+}
+
+pub fn setHeadless(self: *Core, value: bool) void {
+    _ = self;
+    _ = value;
+}
+
+pub fn headless(_: *Core) bool {
+    unreachable;
```
slimsag, line 209: should return `false` not be unreachable

src/platform/wasm/Core.zig
```diff
@@ -0,0 +1,297 @@
...
+}
+
+pub fn setVSync(self: *Core, mode: VSyncMode) void {
+    _ = self;
+    _ = mode;
+    unreachable;
+}
+
+pub fn vsync(_: *Core) VSyncMode {
+    unreachable;
```
slimsag, line 219: We should add a TODO here noting: https://github.com/gpuweb/gpuweb/issues/1224

But I think `unreachable` is the wrong behavior for no, we should just return the "vsync on" mode enum.

slimsag:
@alichraghi do you have a corresponding update to mach-examples? Once that is up I will merge both.

src/platform/wasm/Core.zig
alichraghi: all of those `unreachable` and `undefined`'s were temporary and i forgot to fix them

alichraghi:
yes. `mach-example` PR is also ready. i will sent it once they represent https://github.com/hexops/mach/pull/664#discussion_r1067808608

alichraghi:
https://github.com/hexops/mach-examples/pull/26

slimsag:
wasmserve appears to be broken:

image

alichraghi:
CI Zig version needs to be updated. also i haven't looked closely into *Package Manager MVP* but there might be some changes too remove that parts.
btw: https://github.com/machlibs/imgui/pull/11

PR #672, opened 2023-01-13
gpu: update interface to latest webgpu headers

alichraghi:
wrench PR needs to be merged + updating dawn revision in `sdk.zig`
https://github.com/hexops/sdk-linux-x86_64/pull/6
https://github.com/hexops/sdk-linux-aarch64/pull/4


libs/gpu-dawn/sdk.zig
```diff
@@ -277,9 +277,6 @@ pub fn Sdk(comptime deps: anytype) type {
             step.addIncludePath(include_dir);
             step.addIncludePath(sdkPath("/src/dawn"));
 
-            if (isLinuxDesktopLike(step.target_info.target)) {
-                step.linkSystemLibraryName("X11");
```
slimsag, line 281: @alichraghi I saw you removed this, are you sure it's ok to remove this?

libs/gpu-dawn/sdk.zig
```diff
@@ -539,15 +536,22 @@ pub fn Sdk(comptime deps: anytype) type {
...
+                .openbsd,
+                .dragonfly,
+                => true,
+                else => false,
+            };
         }
 
         pub fn appendFlags(step: *std.build.LibExeObjStep, flags: *std.ArrayList([]const u8), debug_symbols: bool, is_cpp: bool) !void {
-            if (debug_symbols) try flags.append("-g1") else try flags.append("-g0");
+            if (!debug_symbols) try flags.append("-g1") else try flags.append("-g0"); // !!!TODO
```
slimsag, line 552: `!!!TODO` ?

libs/gpu-dawn/sdk.zig
```diff
@@ -564,14 +568,11 @@ pub fn Sdk(comptime deps: anytype) type {
...
 
-            // TODO(build-system): pass system SDK options through
-            try deps.glfw.link(b, lib, .{ .system_sdk = .{ .set_sysroot = false } });
-
             var cpp_flags = std.ArrayList([]const u8).init(b.allocator);
             try appendFlags(step, &cpp_flags, options.debug, true);
             try appendDawnEnableBackendTypeFlags(&cpp_flags, options);
             try cpp_flags.appendSlice(&.{
-                include(deps.glfw_include_dir),
+                "-I" ++ deps.glfw_include_dir,
```
slimsag, line 575: Why like this instead of using `include()`?

libs/gpu-dawn/sdk.zig
```diff
@@ -1303,12 +1321,11 @@ pub fn Sdk(comptime deps: anytype) type {
                     separate_lib.install();
                 break :blk separate_lib;
             };
-            try deps.glfw.link(b, lib, .{ .system_sdk = .{ .set_sysroot = false } });
 
             var flags = std.ArrayList([]const u8).init(b.allocator);
             try appendDawnEnableBackendTypeFlags(&flags, options);
             try flags.appendSlice(&.{
-                include(deps.glfw_include_dir),
+                "-I" ++ deps.glfw_include_dir,
```
slimsag, line 1328: same question

libs/gpu/src/types.zig
```diff
@@ -519,11 +533,13 @@ pub const Limits = extern struct {
     min_uniform_buffer_offset_alignment: u32 = limit_u32_undef,
     min_storage_buffer_offset_alignment: u32 = limit_u32_undef,
     max_vertex_buffers: u32 = limit_u32_undef,
+    max_buffer_size: u32 = limit_u32_undef,
```
slimsag, line 536: This should be a u64 according to the diff:

```diff
     uint32_t maxVertexBuffers;
+    uint64_t maxBufferSize;
     uint32_t maxVertexAttributes;
```

src/platform/native/Core.zig
```diff
@@ -116,7 +116,7 @@ pub fn init(allocator: std.mem.Allocator, options: Options) !*Core {
     }
 
     // Print which adapter we are going to use.
-    var props: gpu.Adapter.Properties = undefined;
+    var props = std.mem.zeroes(gpu.Adapter.Properties);
```
slimsag, line 119: Why would this change? Are we sure that zero values is correct and not `.{}` (default values)?

libs/gpu/src/dawn.zig
```diff
@@ -57,3 +59,19 @@ pub const TogglesDeviceDescriptor = extern struct {
...
+    },
+    power_preference: PowerPreference = .undefined,
+};
+
+const BufferDescriptorErrorInfoFromWireClient = extern struct {
+    chain: ChainedStruct = .{
+        .next = null,
+        .s_type = .dawn_buffer_descriptor_error_info_from_wire_client,
+    },
+    out_of_memory: bool = false,
```
slimsag, line 76: We can see in [dawn.json `"chain roots": ["buffer descriptor"],`](https://github.com/hexops/dawn/blob/generated-2023-01-12.1673546526/dawn.json#L3022-L3025) which tells us this is a valid type to specify in `gpu.Buffer.Descriptor.next_in_chain`, so we need to update `gpu.Buffer.Descriptor.next_in_chain` data type to be an `extern union` where you can specify this type, similar to what we do for `Device.Descriptor.next_in_chain`:

https://github.com/hexops/mach/blob/c02578721a2aa695a7c80802ab8903d4c3f75e41/libs/gpu/src/device.zig#L46-L52

libs/gpu-dawn/sdk.zig
alichraghi: because `include()` make it relative to `/libs/gpu-dawn/`. so you can't e.g give it `/libs/glfw/upstream/glfw/include` from `/build.zig`

libs/gpu/src/dawn.zig
```diff
@@ -57,3 +59,19 @@ pub const TogglesDeviceDescriptor = extern struct {
         };
     }
 };
+
+const AdapterPropertiesPowerPreference = extern struct {
```
slimsag, line 63: Need to do the same thing described in https://github.com/hexops/mach/pull/672/files#r1070298339

dawn.json says:

https://github.com/hexops/dawn/blob/d8468ba09e23c482de1012e00b3adcf98dfe950f/dawn.json#L3013-L3021

Note that this is a chained `out` struct, and goes in `"chain roots": ["adapter properties"],` - so it corresponds to `gpu.Adapter.Properties.next_in_chain`

src/platform/native/Core.zig
alichraghi: dawn now verifies next chain is `null`.
```zig
var props: gpu.Adapter.Properties = undefined;
props.next = null;
```
will also work

slimsag: I think we should use `var props: gpu.Adapter.Properties = .{};` because then we would get the default values (and should get `.next = null`)

slimsag: Ahh I see, if we did `.{}` default values we would need to specify `undefined` for all other fields, so `std.mem.zeroes` is easiest/cleanest. Makes sense.

PR #673, opened 2023-01-15
gpu: add glfw dependency to example and fix issue

rudedogg:
I noticed the triangle example in `libs/gpu` stopped working (ran with `zig build run-example`).

It looks like https://github.com/hexops/mach/commit/a52c6e5f5c92ae0f83f919dd6fd088db0dce34db accidentally removed glfw from `libs/gpu/build.zig` and `libs/gpu/sdk.zig`. I could be missing something, but re-adding the dependency got the `libs/gpu` example building again.

One other issue was the change to Dawn that requires the next chain to be null needed handled. I blindly copied the change discussed here:  https://github.com/hexops/mach/pull/672#pullrequestreview-1248810820

### DIFF
--- a/libs/gpu/build.zig
+++ b/libs/gpu/build.zig
@@ -29,6 +29,7 @@ pub fn build(b: *std.build.Builder) !void {
     example.addPackage(gpu.pkg);
     example.addPackage(glfw.pkg);
     try gpu.link(b, example, .{ .gpu_dawn_options = gpu_dawn_options });
+    try glfw.link(b, example, .{});
     example.install();
 
     const example_run_cmd = example.run();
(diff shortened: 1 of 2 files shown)


rudedogg:
Shoot, looks like this fixes the gpu example but breaks other stuff.

alichraghi:
@rudedogg glfw linkage isn't accidentally removed from `libs/gpu/sdk.zig`. i think the proper fix is linking glfw directly to `gpu` (not `gpu-dawn`) with `glfw.link(b, example, .{})`

rudedogg:
> @rudedogg glfw linkage isn't accidentally removed from `libs/gpu/sdk.zig`. i think the proper fix is linking glfw directly to `gpu` (not `gpu-dawn`) with `glfw.link(b, example, .{})`

Ah I see. Thanks!

I force-pushed the fix you suggested, now mach itself builds, and the gpu example triangle still works. 

Let me know if anything else needs changed.

alichraghi (approved):

slimsag (approved):
Thank you!

PR #681, opened 2023-01-17
mach: eliminate heap allocation of Core

slimsag:
Prior to this change `mach.Core.init` would heap allocate the structure, returning `*Core`:

```zig
app.core = try mach.Core.init(allocator, .{});
```

This was obviously not ideal, but wasn't possible to eliminate before due to how Core was entangled with the platform abstraction. Now that it has been removed, we can reduce Core initialization to take a `*Core` to initialize. In practice this means initialization looks something like this:

```zig
try mach.Core.init(&app.core, alloctor, .{});
```

Or more simply:

```zig
try app.core.init(allocator, .{});
```

And we eliminate the `*Core` allocation entirely in most cases.

### DIFF
--- a/src/platform/wasm/Core.zig
+++ b/src/platform/wasm/Core.zig
@@ -25,7 +25,7 @@ id: js.CanvasId,
 last_cursor_position: Position,
 last_key_mods: KeyMods,
 
-pub fn init(core: *Core, allocator: std.mem.Allocator, options: Options) !*Core {
+pub fn init(core: *Core, allocator: std.mem.Allocator, options: Options) !void {
     _ = options;
     var selector = [1]u8{0} ** 15;
     const id = js.machCanvasInit(&selector[0]);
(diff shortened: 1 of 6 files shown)


slimsag:
See also https://github.com/hexops/mach-examples/pull/29

@alichraghi thoughts?

src/platform/wasm/Core.zig
```diff
@@ -25,14 +25,12 @@ id: js.CanvasId,
 last_cursor_position: Position,
 last_key_mods: KeyMods,
 
-pub fn init(allocator: std.mem.Allocator, options: Options) !*Core {
+pub fn init(core: *Core, allocator: std.mem.Allocator, options: Options) !*Core {
```
alichraghi, line 28: this still returns `*Core`

alichraghi (changes requested):

alichraghi:
i approve this :+1:

PR #694, opened 2023-01-24
all: split Mach into Core and Engine parts

slimsag:
This change splits Mach `src/` into two separate codebases proper, so that we can really start developing higher level apps.

* `libs/core` - Mach _core_, a standalone small library which provides a modern alternative to SDL/etc. It provides a Window+Input+GPU, nothing else.
* `src/` - Mach _engine_, where higher-level Mach code will live, ECS modules, rendering facilities, etc.


slimsag:
cc @alichraghi

PR #695, opened 2023-01-25
core: make Core.pollEvents return an iterator; remove Core.hasEvent

slimsag:
After this change:
    
* `Core.pollEvents` returns an iterator. At the time of polling events, Mach core will perform work to poll for events, handle resizing of the framebuffer, etc. and the iterator allows the caller to consume all available events.
* The event queue is now baced by a `std.fifo.LinearFifo`, which removes the need for dynamic allocation of each event. Instead, the event queue starts with a generous size suitable for most high-end gaming setups (high-precision mouse, etc.) and can grow, but never shrink, up to the maximum event queue size experienced by the app within any given frame. Effectively, this means we find the maximum capacity needed to store events and avoid runtime allocations.
* `Core.hasEvent` is removed.

Notably, since event polling on Desktop platforms is now done inside of `Core.pollEvents` instead of once-per-call to `pollEvents`, this is much faster. It additionally still allows for one to poll for events multiple times per frame if you like, for handling events mid-frame if you desire.

### DIFF
--- a/libs/core/src/Core.zig
+++ b/libs/core/src/Core.zig
@@ -24,12 +24,16 @@ pub fn deinit(core: *Core) void {
     return core.internal.deinit();
 }
 
-pub fn hasEvent(core: *Core) bool {
-    return core.internal.hasEvent();
-}
+pub const EventIterator = struct {
+    internal: platform.Core.EventIterator,
 
-pub fn pollEvents(core: *Core) ?Event {
-    return core.internal.pollEvents();
+    pub inline fn next(self: *EventIterator) ?Event {
+        return self.internal.next();
+    }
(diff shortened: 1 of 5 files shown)


slimsag:
cc @alichraghi

slimsag:
note: I plan to make another change soon which undoes the rename and makes `pollEvents` return an iterator, this way we can poll for events without even rendering a frame - or poll twice in one frame. Important for handling events mid-frame if one wishes to do so.

slimsag:
I just pushed a much better approach, which makes us use an iterator type instead (and has us poll for GLFW events when grabbing the iterator)

PR #696, opened 2023-01-28
gpu-dawn: update gzip decompression to use latest stdlib API

kimshrier:
A simple rename of a function from the standard library.


slimsag (approved):
Awesome, than you!

slimsag:
@rob-bateman what `zig version` are you using? It sounds like you may not be using the more recent (but not absolute latest) version we support, i.e. https://github.com/hexops/mach#supported-zig-version

rob-bateman:
I'm using 0.11.0-dev.1414. It is probably out of date now but ti was bleeding edge when i downloaded (ie not the stable 0.10.x release) so it would be surprising for that to be the issue, but I'll update and try again. Thx!

PR #699, opened 2023-01-31
gpu-dawn: fix git error when fetching Dawn source

Luexa:
Previously, when the gpu-dawn SDK attempted to fetch Dawn sources, Git failed with this error:

```
fatal: ambiguous argument 'generated-*': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

This caused the whole build to fail as the gpu-dawn SDK is unable to check out the intended revision of the generated Dawn sources. This commit fixes the issue by using `git checkout --force` instead of `git reset --hard`, which works because you can checkout to a detached commit hash (used by DirectXShaderCompiler, for example) as well as an actual branch.

The alternative would be to do `git reset --hard origin/generated-*` rather than `git reset --hard generated-*`, but that would require modifying Wrench unlike this solution.


slimsag (approved):
Thanks for the fix! and sorry about introducing this bug, I'd meant to look into it after you mentioned it but lost track of it :)

PR #717, opened 2023-03-10
core: fix usage of std_options for log_level and log_scope_levels

dweiller:
This commit fixes exporting of the logging std lib options, which was not updated for the `std_options` change.


slimsag:
@dweiller BTW I forget, are you in discord? I was keen to ask you how the app work is going but couldn't find you :)

PR #738, opened 2023-04-02
app: add mach CLI and integrate wasmserve

### LINKED ISSUE #595: mach: sort out CLI arguments / program options
* [ ] If one follows the [getting started guide](https://github.com/hexops/mach-examples#use-mach-engine-in-your-own-project), it's not possible to easily modify the `build.zig` to pass arguments from `zig build run -- foobar` to your program.
    * i.e. you can't call `.addArgs` because `getInstallStep` returns a `Step` not a `RunStep`.
    * We cannot return a `RunStep`, since for WASM targets the step runs `wasmserve` so you can open in your browser.
    * We may need to expose a separate function like `runNative` (squirl's suggestion), or come up with a different solution.
    * WASM (browser) targets, and likely also other targets (mobile, VR, etc.) do not really have a concept of CLI flags. We may need to come up with some other form of simple "options" one can pass to their program that works in all these situations nicely.
* [ ] Whatever we come up with, it should have a documentation section on the new machengine.org website: "adding CLI arguments" or similar.

### LINKED ISSUE #707: wasmserve: major regressions with latest Zig build changes (wasmserve is broken)
Talking with @alichraghi - wasmserve has a copy of a big stdlib build function, which is no longer up to date with the Zig build changes. In [this PR](https://github.com/hexops/mach/pull/706/files#diff-b91a16298446887ba7580572619d5ad30b19585a5ab04124909e245e2488e4af) I removed the function, effectively breaking wasmserve.

Context: since we are running `wasmserve` as a library in `zig build` the only way is using `zig [commands]` or `zig build [step]` (the old way)

We have two paths forward:

1. Propose exposing command creator in stdlib so we can use it without copying it.
2. Use `zig build [step]` instead, and require that projects using Mach/wasmserve expose a "known" build step; i.e. what should `wasmserve` run when building the project?

I think #2 is possibly the better path forward here. In theory, the user can pass in the `[step]` alone as an argument to Mach/wasmserve, and then wasmserve should merely invoke `zig build [step] -Dtarget=wasm32-freestanding`?

alichraghi:
this is how you e.g. run sysaudio example in [mach-examples](https://github.com/hexops/mach-examples)
```
mach build sysaudio --serve 
```

See Also [mach-examples/#47](https://github.com/hexops/mach-examples/pull/47)

Closes #707, Helps #595


slimsag (approved):

slimsag:
I'm not sure this fixes #595 exactly - part of that issue is having some way to specify arguments to e.g. `mach build`, have the example launch/run, and have those arguments get passed to the user's program - including in the browser. Does this achieve it somehow?

alichraghi:
yep.
```
mach build run -- -- custom_arg
```

PR #748, opened 2023-04-08
harfbuzz: expose C API

erikarvstedt:
Like in `freetype/src/main.zig`.


slimsag (approved):
Not opposed to this; but is there anything we can do nicer in the Zig API so you don't need this?

erikarvstedt:
I needed it for a minor size-optimization: `@sizeOf(?harfbuzz.Buffer)`  is 2x larger than `@sizeOf(?*harfbuzz.c.hb_buffer_t)`.
Hopefully, this won't be needed in future Zig versions.

Issue #622, opened 2022-11-22
editor: wasm serving: watch file system with `std.fs.Watch`


alichraghi:

alichraghi:
macos has a whole different API to watch file system changes :( https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html

PR #750, opened 2023-04-10
app: Use std.ChildProcess instead of POSIX calls

### LINKED ISSUE #739: app: windows target CI failure
https://github.com/hexops/mach/actions/runs/4590437389/jobs/8105981801

tato:
fixes #739

In my testing it builds like in the fork/exec implementation. But I couldn't get any of the examples to run under wasm, even under the old implementation.

### DIFF
--- a/app/Builder.zig
+++ b/app/Builder.zig
@@ -34,13 +34,15 @@ const Status = union(enum) {
 };
 
 pub fn run(self: *Builder) !void {
+    var child = try self.runZigBuild(.Inherit);
+    switch (try child.wait()) {
+        .Exited => |code| {
+            if (code != 0) std.os.exit(code);
+        },
+        else => std.os.exit(1),
+    }
+
     if (self.serve) {
-        const child_pid = try std.os.fork();
-        if (child_pid == 0) try self.exec();
-
-        const wait_res = std.os.waitpid(child_pid, 0);
-        if (wait_res.status != 0) std.os.exit(1);
-
         var out_dir = std.fs.cwd().openIterableDir(out_dir_path, .{}) catch |err| {
             std.log.err("cannot open '{s}': {s}", .{ out_dir_path, @errorName(err) });
             std.os.exit(1);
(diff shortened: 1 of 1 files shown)


app/Builder.zig
```diff
@@ -88,48 +84,42 @@ pub fn run(self: *Builder) !void {
...
+
+    const args = try self.buildArgs(args_arena.allocator());
 
-    return std.os.execvpeZ(
-        argv[0].?,
-        @ptrCast([*:null]const ?[*:0]const u8, argv),
-        @ptrCast([*:null]const ?[*:0]const u8, std.os.environ.ptr),
-    );
+    var child = std.ChildProcess.init(args, allocator);
+    _ = try child.spawnAndWait();
```
alichraghi, line 97: doesn't check exit status

alichraghi:
Hi @tato. thanks for working on this. i just tested it and it works fine on linux.
you can test it using:
(under mach-examples repo)
```
<PATH_TO_MACH_REPO>/zig-out/bin/mach build sysaudio --serve
```

app/Builder.zig
tato: True 😅 I guess it should just exit if the result is == .Signal / .Stopped / .Unknown too?

alichraghi: exactly

alichraghi (approved):

alichraghi:
this doesn't close #739 tho. because there's another fork in https://github.com/hexops/mach/blob/1841ed056d41dd19b0a7d238a52b1279547b6477/app/Builder.zig#L323 which can't be solved easily since `pipe` won't work this way so we can't collect stderr

tato:
Should be solvable with std.ChildProcess too in a cross-platform way. I didn't realise there was another `fork` call 😅

alichraghi:
hmm not sure. needs testing. also `std.ChildProcess` has been renamed to `std.process.Child`

alichraghi:
https://ziglang.org/documentation/master/std/src/child_process.zig.html#L48

alichraghi:
nothing yet. watcher will call it once #622 get fixed. also LGTM (this fixes #739)

slimsag (approved):

PR #770, opened 2023-05-07
mach: fix deprecated mem.set() call

grantmor:
Update to the @memset builtin fixes the cubemap and pbr-basic examples assuming zigimg is also up to date

mem.set() appears to have been deprecated

This change gets the cubemap and pbr-basic demos running again with zig11-dev

The zigimg dependency also calls the mem.set() function, and it looks like they have already updated on their end, but I'm not experienced enough with the codebase or github to know how that would get pulled in

This is my first pull request so I'm not sure if I'm following all the correct procedures here

Let me know if any additional changes are necessary :)


slimsag (approved):
Sorry for the delay, and thank you!

PR #781, opened 2023-05-22
gpu: fix few chained struct wrappers

hi-void:
Changed `InstanceDescriptor` and `TogglesDeviceDescriptor` structs to make them compatible with the definitions in Dawn's `webgpu.h`. Before this, it wasn't possible to provide alternative library search paths for Dawn's Vulkan backend or toggle Dawn-specific feature flags like enabling the usage of unsafe or experimental WebGPU extensions. Namely the `timestamp-query` extension, which is disabled on the grounds of security as it exposes precise hardware timings that may be used to track users on the Web, which probably isn't relevant when using WebGPU natively.

The `WGPUDawnTogglesDeviceDescriptor` struct wrapped by `TogglesDeviceDescriptor` is deprecated by Dawn and will be removed in the future. In preparation for that, also added a `TogglesDescriptor` struct wrapping the `WGPUDawnTogglesDescriptor` struct meant to replace it. This struct is a drop-in replacement for `TogglesDeviceDescriptor` in the descriptor struct for `createDevice` and can also be used in the descriptor struct for `createInstance`, but I might've jumped the gun on this because the version of Dawn we use doesn't support its use yet and it does nothing for now.

### DIFF
--- a/libs/gpu/src/dawn.zig
+++ b/libs/gpu/src/dawn.zig
@@ -60,29 +60,6 @@ pub const TogglesDeviceDescriptor = extern struct {
     }
 };
 
-pub const TogglesDescriptor = extern struct {
-    chain: ChainedStruct = .{ .next = null, .s_type = .dawn_toggles_descriptor },
-    enabled_toggles_count: u32 = 0,
-    enabled_toggles: ?[*]const [*:0]const u8 = null,
-    disabled_toggles_count: u32 = 0,
-    disabled_toggles: ?[*]const [*:0]const u8 = null,
-
-    /// Provides a slightly friendlier Zig API to initialize this structure.
-    pub inline fn init(v: struct {
-        chain: ChainedStruct = .{ .next = null, .s_type = .dawn_toggles_descriptor },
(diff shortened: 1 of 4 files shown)


libs/gpu/src/dawn.zig
```diff
@@ -60,6 +60,29 @@ pub const TogglesDeviceDescriptor = extern struct {
     }
 };
 
+pub const TogglesDescriptor = extern struct {
```
slimsag, line 63: Many thanks for this PR; could you remove `TogglesDescriptor` for now, though? I want to make sure that we 1:1 reflect what is in Dawn's `webgpu.h`, when we do update later I can add this back - I just want to ensure we're 1:1 with the header API.

hi-void: That makes sense, done! I also went ahead and renamed the MR because there are no new structs in here anymore.

slimsag (approved):

slimsag:
Thanks again for this! :) Much appreciated

Issue #796, opened 2023-06-03
100% banish usingnamespace from all of our codebases


slimsag:
There is only 1 exception to this: `mach-core/platform` entrypoints have an actual use case for this and we should keep it there.

slimsag:
The remaining users:

- [ ] mach-glfw
- [ ] mach-basisu
- [ ] mach-sysjs (`next` branch, `usingnamespace @import("sysjs_generated.zig");`)
- [ ] mach-freetype

PR #798, opened 2023-06-04
gpu: remove usingnamespace for callbacks.zig and types.zig

tech0tron:
Continuing off of issue #796 and pull request #797, I removed the usingnamespaces for `callbacks.zig` and `types.zig` and added their types.

I wrote up a basic python script for this: it looks for all `pub const`s or `pub fn`s in the file and then generates the re-exports. It only writes one `@import` per each file and then uses a variable to do the re-exports, but I did a simple find and replace for `@import`s on each declaration, to keep it unified with what #797 did.

Here's the script if this could be useful for removing more usingnamespace: https://gist.github.com/tech0tron/e26d54d057e6c030b1d773dcf3b5350c


libs/gpu/src/main.zig
```diff
@@ -28,7 +34,88 @@ pub const TextureView = @import("texture_view.zig").TextureView;
 
 pub const dawn = @import("dawn.zig");
 
-pub usingnamespace @import("types.zig");
+pub const array_layer_count_undef = @import("types.zig").array_layer_count_undef;
```
slimsag, line 37: I mentioned this earlier in Discord:

> my suggestion for callbacks.zig and types.zig is to move all that code directly into main.zig

I think that:

1. The fact that we need to re-export so many declarations, that you felt the need to write a script to do so, is a good hint that something is wrong :)
2. Elevating those declarations right into main.zig eliminates that problem (and the problem of keeping this list of re-exported declarations up to date in the future)
3. Having two files named `types.zig` and `callbacks.zig` is a bit of an anti-pattern in general, anyway - so best we move away from that.

Does that sound reasonable?

tech0tron: Done

tech0tron: Done

slimsag (approved):
Excellent work, thank you!

PR #804, opened 2023-06-09
gpu: make `swapChainGetCurrentTextureView()` return value optional

razcore-rad:

This came up in Discord chat, that `swapChainGetCurrentTextureView()` may fail, but it always tries to return the value even if on the C side the pointer might be `NULL`.

This PR fixes updates the return value of this function to be `optional` so we can check for it in zig. Further context: <https://eliemichel.github.io/LearnWebGPU/getting-started/first-color.html#texture-view>


slimsag (approved):
Thank you!

PR #807, opened 2023-06-12
freetype: fix and re-enable harfbuzz

mlugg:
This wasn't actually broken due to a Zig bug, but rather due to some pragmas in Harfbuzz which forcibly enable this -Werror. Luckily, there's a macro you can define to disable them.


slimsag (approved):
Thanks!

PR #814, opened 2023-06-20
mach: entry fix for overriding logging scope levels and exposed option for custom function

RokKos:
I fixed logging override options in `native/entry.zig` and added functionality to define the logging function.

The reason is that users can define in `main.zig` their own logging scope levels and logging function. For example something along the lines of:

```zig
pub const App = @This();

pub const std_options = struct {
    pub const log_scope_levels = &[_]std.log.ScopeLevel{ 
      .{ .scope = .rendering, .level = .debug }, 
      .{ .scope = .gameplay, .level = .info },
      .{ .scope = .animations, .level = .err } };
    pub const logFn = customLogFn;
};

pub fn customLogFn(
    comptime level: std.log.Level,
    comptime scope: @TypeOf(.EnumLiteral),
    comptime format: []const u8,
    args: anytype,
) void {
    // Custom formating
}
```

I also checked `wasm/entry.zig` and saw that there was no issue and it already has some custom logging functions.

Reference used:
- [zig std.log](https://ziglang.org/documentation/master/std/#A;std:log)
- [Overview of std.log](https://gist.github.com/leecannon/d6f5d7e5af5881c466161270347ce84d)

### DIFF
--- a/libs/core/src/platform/native/entry.zig
+++ b/libs/core/src/platform/native/entry.zig
@@ -17,6 +17,11 @@ pub const std_options = struct {
         app_std_options.log_scope_levels
     else
         &[0]std.log.ScopeLevel{};
+
+    pub const logFn = if (@hasDecl(app_std_options, "logFn"))
+        app_std_options.logFn
+    else
+        std.log.defaultLog;
 };
 
 pub fn main() !void {
--- a/libs/core/src/platform/native/entry.zig
+++ b/libs/core/src/platform/native/entry.zig
@@ -13,7 +13,7 @@ pub const std_options = struct {
     else
         std.log.default_level;
 
-    pub const log_scope_levels = if (@hasDecl(App, "log_scope_levels"))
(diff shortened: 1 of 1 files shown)


slimsag:
@alichraghi can I get your opinion on this code?

I'm also reallyannoyed that we special case std_options here. I wonder why we cannot just re-export whatever is in the user's main file, or at least all `std_options`, without any knowledge of the std log package

slimsag:
After https://github.com/hexops/mach-core/commit/8c91dae246481a6b1dac9b2bcd5589e333047456 it is now possible to access your `main.zig`'s declarations via `@import("root").foo` as you'd expect. Your app's `main.zig` can now declare `std_options` like so:

```
pub const std_options = struct {
    pub const logFn = https://github.com/import("root").machLogFn;

    // ... your custom options ...
};

// ... init/update/deinit functions ...
```

PR #816, opened 2023-06-24
core: add is_headless to specify hidden window before init

undefinedDarkness:
I hope it's no trouble 

### DIFF
--- a/libs/core/src/Core.zig
+++ b/libs/core/src/Core.zig
@@ -8,7 +8,7 @@ pub const Core = @This();
 internal: platform.Core,
 
 pub const Options = struct {
-    is_app: bool = true,
+    is_app: bool = false,
     is_headless: bool = false,
     title: [*:0]const u8 = "Mach Engine",
     size: Size = .{ .width = 1920 / 2, .height = 1080 / 2 },
--- a/libs/core/src/Core.zig
+++ b/libs/core/src/Core.zig
@@ -8,7 +8,8 @@ pub const Core = @This();
 internal: platform.Core,
 
 pub const Options = struct {
-    is_app: bool = false,
+    is_app: bool = true,
+    is_headless: bool = false,
     title: [*:0]const u8 = "Mach Engine",
     size: Size = .{ .width = 1920 / 2, .height = 1080 / 2 },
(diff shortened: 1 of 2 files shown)


slimsag:
Thanks, I'm happy to accept an improvement like this in general but I'm afraid this is not the right API.

Can you describe the issue you want to solve here? Is it:

a. You don't want to have to call `setHeadless` (API is burdensome)?
b. You are seeing the window appear and then disappear (flicker), and need a way to specify headless earlier?

The reason this enum-based API is not great is because one could easily have e.g. a game application, which sometimes may be run `game.exe --headless` - so in that case what should the user pick?

libs/core/src/Core.zig
```diff
@@ -7,8 +7,10 @@ pub const Core = @This();
 
 internal: platform.Core,
 
+pub const AppType = enum { Standard, Game, Headless };
+
 pub const Options = struct {
-    is_app: bool = false,
+    app_type: AppType = .Game,
```
slimsag, line 13: I suspect the solution here will be to:

1. Keep `is_app: bool = false,`
2. Additionally add `headless: bool = false,` (below `size` field) which we can respect at `init`

undefinedDarkness: Sure, that works as well

slimsag: cool, if you're happy with that I'm happy

want to send a new PR or update this one? or prefer if I make that change? either way is fine with me

undefinedDarkness:
> You are seeing the window appear and then disappear (flicker), and need a way to specify headless earlier?
Yeah just this bit, the rest doesn't matter, The window stays open till core initiates which takes a few,

libs/core/src/Core.zig
```diff
@@ -8,7 +8,8 @@ pub const Core = @This();
 internal: platform.Core,
 
 pub const Options = struct {
-    is_app: bool = false,
+    is_app: bool = true,
```
slimsag, line 11:
```suggestion
    is_app: bool = false,
```

slimsag (approved):

slimsag:
Thanks!

PR #914, opened 2023-08-03
update to latest mach-core API

alichraghi:

### DIFF
--- a/src/editor/app.zig
+++ b/src/editor/app.zig
@@ -1,8 +1,10 @@
 const std = @import("std");
 const mach = @import("mach");
+const core = mach.core;
 const gpu = mach.gpu;
 
-pub const App = @This();
+pub const name = .editor;
+pub const App = mach.App(.{ mach.Module, @This() });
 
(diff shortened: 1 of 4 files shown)


src/editor/app.zig
```diff
@@ -1,8 +1,10 @@
 const std = @import("std");
 const mach = @import("mach");
+const core = mach.core;
 const gpu = mach.gpu;
 
-pub const App = @This();
+pub const name = .editor;
+pub const App = mach.App(.{ mach.Module, @This() });
```
slimsag, line 7: Passing `@This()` into `mach.App` as a module is.. sketchy and seems very wrong. Why is this necessary?

slimsag: I get what you're doing - fine for now, but let's clean this up later.

slimsag (approved):

src/editor/app.zig
alichraghi: otherwise, init, deinit and update can't be called

PR #1028, opened 2023-09-23
math: Rewrite Vec tests and fix Vec cross calculation bug

noisegul:

---

Rewrote the old commented out tests in accordance to the Mach style guide and added the listed missing basic tests.
While doing so I also caught two bugs in the cross and swizzling methods:
- the naming of the Vec3 swizzling method was wrong, using `yzw` instead of `yzx`, and `.w()` instead of `.x()`, while cross was calling `yzx`
- the cross method used e.g. `.mul(b.xzy())` instead of `.mul(&b.xzy())`

### DIFF
--- a/src/math/vec.zig
+++ b/src/math/vec.zig
@@ -512,16 +512,167 @@ test "minScalar" {
     try testing.expect(f32, -182).eql(math.Vec4.minScalar(&b, &c));
 }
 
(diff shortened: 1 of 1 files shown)


slimsag (approved):
This is super, super helpful - thank you!

PR #1032, opened 2023-09-23
math: Implement new Vec swizzle method

noisegul:
---
Motivated by the previously associated TODO note.
This strategy allows for generalized arbitrary swizzling instead by using an enum and the builtin [\@shuffle](https://ziglang.org/documentation/master/#shuffle) function.
- old: `a.yzx()`
- new: `a.swizzle(.y, .z, .x)`


slimsag (approved):
Thanks @noisegul - this looks like a good approach

Issue #1004, opened 2023-09-16
math: add Vec*Mat and Mat*Vec operations


slimsag:
* [x] Mat*Vec https://github.com/hexops/mach/pull/1051
* [ ] Vec*Mat TODO

slimsag:
Fixed by https://github.com/hexops/mach/pull/1054

PR #1051, opened 2023-10-03
Added mulVec function for matrices

omprakaash:
Added function to multiply a matrix with a vector (Matrix* Vec).

### DIFF
--- a/src/math/mat.zig
+++ b/src/math/mat.zig
@@ -360,7 +360,7 @@ pub fn Mat(
             return result;
         }
 
-        /// Matrix - Vector multiplicatino: Mat * Vec
+        /// Matrix * Vector multiplication
         pub inline fn mulVec(a: *const Matrix, b: *const ColVec) ColVec {
             var result = [_]ColVec.T{0}**ColVec.n;
             inline for (0..Matrix.rows) |row| {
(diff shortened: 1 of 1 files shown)


src/math/mat.zig
```diff
@@ -636,3 +652,39 @@ test "Mat4x4_perspective" {
...
+        &math.vec3(0, 2, 0),
+        &math.vec3(0, 0, 3),
+    );
+
+    const m = math.Mat3x3.mulVec(&mat, &v);
+    const expected = math.vec3(2,2,3);
+    try testing.expect(math.Vec3, expected).eql(m);
+}
+
+test "Mat4x4_mulVec_ve4" {
```
slimsag, line 678:
```suggestion
test "Mat4x4_mulVec_vec4" {
```

src/math/mat.zig
```diff
@@ -357,6 +360,19 @@ pub fn Mat(
             return result;
         }
 
+        /// Matrix - Vector multiplicatino: Mat * Vec
```
slimsag, line 363:
```suggestion
        /// Matrix * Vector multiplication
```

slimsag (approved):
Thanks for sending this!

slimsag:
@omprakaash would you be interested in adding Vec*Mat (the opposite order)? I think that's all we need to close https://github.com/hexops/mach/issues/1004

omprakaash:
> @omprakaash would you be interested in adding Vec*Mat (the opposite order)? I think that's all we need to close #1004

Sure ! Will get started pretty soon.

PR #1054, opened 2023-10-04
Add mulVec function for vectors

omprakaash:
Add function to allow a matrix to be left multiplied by a vector. Not quite sure about the naming of the multiplication functions. Would be happy to hear any suggestions if needed.  

### DIFF
--- a/src/math/mat.zig
+++ b/src/math/mat.zig
@@ -361,15 +361,16 @@ pub fn Mat(
         }
 
         /// Matrix * Vector multiplication
-        pub inline fn mulVec(a: *const Matrix, b: *const ColVec) ColVec {
-            var result = [_]ColVec.T{0} ** ColVec.n;
+        pub inline fn mulVec(matrix: *const Matrix, vector: *const ColVec) ColVec {
+            var result = [_]ColVec.T{0}**ColVec.n;
             inline for (0..Matrix.rows) |row| {
                 inline for (0..ColVec.n) |i| {
-                    result[i] += a.v[row].v[i] * b.v[row];
+                    result[i] += matrix.v[row].v[i] * vector.v[row];
                 }
             }
             return vec.Vec(ColVec.n, ColVec.T){ .v = result };
         }
+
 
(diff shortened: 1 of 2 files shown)


src/math/mat.zig
```diff
@@ -361,11 +361,24 @@ pub fn Mat(
...
+                    result[i] += matrix.v[row].v[i] * vector.v[row];
+                }
+            }
+            return vec.Vec(ColVec.n, ColVec.T){
+                .v = result
+            };
+        }
+
+        /// Vector * Matrix multiplication
+        pub inline fn mulVecL(vector: *const RowVec, matrix: *const Matrix) ColVec{
```
slimsag, line 377: I think this is the wrong place to have this code live (which also explains the naming issue you raised):

* `mat.mulVec(vec)` should multiply a `Matrix * Vector`, which you already have implemented
* `vec.mulMat(mat)` should multiply a `Vector * Matrix`, which you are implementing in this PR currently as `mulVecL`.

i.e. this code and the tests could be moved directly into `vec.zig`

src/math/vec.zig
```diff
@@ -70,6 +71,20 @@ pub fn Vec(comptime n_value: usize, comptime Scalar: type) type {
                         .mul(&b.swizzle(.y, .z, .x));
                     return s1.sub(&s2);
                 }
+
+                /// Vector * Matrix multiplication
+                pub inline fn mulMat(vector: *const VecN, matrix: *const mat.Mat(3, 3, Vec(4, T) )) VecN{
```
slimsag, line 76:
```suggestion
                pub inline fn mulMat(vector: *const VecN, matrix: *math.Mat3x3) VecN{
```

Could we write the type name in these instead so it's easier for users to make sense of it?

src/math/mat.zig
omprakaash: Ah got you. Moved function to` vec.zig` .

slimsag (approved):

PR #1059, opened 2023-10-07
math: More precise triangleIntersect tests and Ray type improvements/bugfixes

noisegul:
-------
Added some more precise test cases for the ray-triangle intersection function.
I created these test cases the following way:
- Generate a random 3D triangle
- Sample a random point on/inside the triangle
- Calculate the barycentric coordinates of the sampled point
- Create a random ray origin point
- Calculate the normalized direction vector
- Calculate distance

(Did the calculations in Mathematica)

------
This PR also fixes a bug with the Ray type so that the `Rayd` and `Rayh` types work properly now and cleaned up the code a bit in the process: e.g. removed the `floatFallbackPrecision` function and replaced it with a simple switch expression. `Ray.Hit` is also now properly generic based on the specific `Ray` type, and will be either a `math.Vec4`, `math.Vec4d` or `math.Vec4h` respectively. In other words: `Ray.Hit` will be `math.Vec4`, `Rayd.Hit` will be `math.Vec4d` etc.

### DIFF
--- a/src/math/ray.zig
+++ b/src/math/ray.zig
@@ -59,7 +59,9 @@ pub fn Ray(comptime Vec3P: type) type {
                 /// Check for collision of a ray and a triangle in 3D space.
                 /// Triangle winding, which determines front- and backface of
                 /// the given triangle, matters if backface culling is to be
-                /// enabled. Without backface culling it does not matter.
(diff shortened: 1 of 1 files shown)


slimsag (approved):
Awesome work! This looks great :) Thanks again for contributing this!

PR #1061, opened 2023-10-09
Add fromInt constructor helper for vectors

omprakaash:
Add fromInt constructor helper for vectors. 


src/math/vec.zig
```diff
@@ -38,6 +41,9 @@ pub fn Vec(comptime n_value: usize, comptime Scalar: type) type {
                 pub inline fn init(xs: Scalar, ys: Scalar, zs: Scalar) VecN {
                     return .{ .v = .{ xs, ys, zs } };
                 }
+                pub inline fn fromInt(xs: i128, ys: i128, zs: i128) VecN {
```
slimsag, line 44: @omprakaash by using `i128` here I think we may be forcing the compiler to use more memory, or at the very least relying on the optimizer to notice that we aren't doing that.

A better approach here might be to use `anytype` instead of `i128`, then anything that is an integer can be passed in and `@floatFromInt` will act however it normally acts when an arbitrary type is given to it.

src/math/vec.zig
```diff
@@ -910,3 +917,30 @@ test "Mat4x4_mulMat" {
     const expected = math.vec4(7, 9, 6, -1);
     try testing.expect(math.Vec4, expected).eql(m);
 }
+
+test "Vec2_fromInt" {
+    const x: i8 = 0;
+    const y: i32 = 1;
+    const v = math.Vec2.fromInt(x, y);
```
slimsag, line 924: Just like how we have `math.vec2(x, y)` which is a helper/alias for `math.Vec2.init(x, y)`... could you add a helper/alias for `math.vec2FromInt(x, y)` and use it in these tests instead? You'd write that here: https://github.com/hexops/mach/blob/fc4c3d06a306f49f01db347e6b1fb8561fb10c72/src/math/main.zig#L76-L98

src/math/vec.zig
omprakaash: Ah got you. Was not thinking about `anytype`. Will change the signature to include it.

src/math/vec.zig
omprakaash: Oh Yeah. Will add those in main.zig

slimsag (approved):
Great work, thanks!

PR #1065, opened 2023-10-11
math: fix clip space of perspective matrix 

slimsag:
The prior implementation https://github.com/hexops/mach/pull/1002 was wrong as it was using clip space [-1, 1] which matches OpenGL but not WebGPU. Our math library is WebGPU / modern API focused, so this corrects that and assumes clip space [0, 1].


slimsag:
Closing for now and will reopen later. I need to:

1. Investigate what pdoane mentioned (was wgpu-matrix wrong?)
2. Implement reverse-z instead and mandate that, https://github.com/hexops/mach/issues/1086

PR #1066, opened 2023-10-12
math: Add Matrix multiplication tests

noisegul:
--------
Motivated by the associated `TODO` comment.
Tests for Matrix multiplication with `Mat4x4` and `Mat3x3`. Caught a bug which causes the `Mat3x3` test to fail. The reason being that `Mat3x3` uses `Vec4`s, I assume this is for padding and compatibility reasons.
The culprit can be found in this line of the `mul` implementation:
```zig
var result: Matrix = undefined;
```
This does not matter for `Mat4x4` since every row and column gets touched. But for when creating a `Mat3x3` we call `init` with a `Vec3` but the `init` method sets the last column of the internal `Vec4` to `1`, so if the values start `undefined` it makes testing against and comparison with an expected `Mat3x3` fail.

Example:

```
============ expected this output: =============  len: 3 (0x3)

[0]: math.vec.Vec(4,f32){ .v = { 4.1e+01, 1.13e+02, 6.7e+01, 1.0e+00 } }
[1]: math.vec.Vec(4,f32){ .v = { -2.2e+01, -4.4e+01, 1.5e+01, 1.0e+00 } }
[2]: math.vec.Vec(4,f32){ .v = { -4.0e+01, -7.0e+01, -8.0e+00, 1.0e+00 } }

============= instead found this: ==============  len: 3 (0x3)

[0]: math.vec.Vec(4,f32){ .v = { 4.1e+01, 1.13e+02, 6.7e+01, -3.03164882e-13 } }
[1]: math.vec.Vec(4,f32){ .v = { -2.2e+01, -4.4e+01, 1.5e+01, -3.03164882e-13 } }
[2]: math.vec.Vec(4,f32){ .v = { -4.0e+01, -7.0e+01, -8.0e+00, -3.03164882e-13 } }

================================================
``` 

Since there are multiple places in which this problem can be fixed, e.g. in the `mul` method itself or by making testing aware of `Mat3x3` and ignore the last column, I have not yet done so. But if you tell me how you want it handled I can update this PR, or send another one.

### DIFF
--- a/src/math/mat.zig
+++ b/src/math/mat.zig
@@ -340,7 +340,6 @@ pub fn Mat(
         };
 
         /// Matrix multiplication a*b
-        // TODO: needs tests
         pub inline fn mul(a: *const Matrix, b: *const Matrix) Matrix {
             @setEvalBranchQuota(10000);
             var result: Matrix = undefined;
@@ -362,7 +361,7 @@ pub fn Mat(
 
         /// Matrix * Vector multiplication
         pub inline fn mulVec(matrix: *const Matrix, vector: *const ColVec) ColVec {
-            var result = [_]ColVec.T{0}**ColVec.n;
+            var result = [_]ColVec.T{0} ** ColVec.n;
             inline for (0..Matrix.rows) |row| {
                 inline for (0..ColVec.n) |i| {
                     result[i] += matrix.v[row].v[i] * vector.v[row];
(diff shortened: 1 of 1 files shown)


slimsag:
@noisegul thanks a bunch for sending / catching this, much appreciated. apologies I just now had a chance to look.

I think we can add `@align()` to the `v: Vector` fields to get proper alignment. I will try this out soon, which I think should help in fixing this?

noisegul:
No problem!

PR #1068, opened 2023-10-16
engine: release encoder properly

Vulfox:
Noticed a memory leak with the basic triangle example. An encoder release was forgotten.


slimsag:
Thanks!

PR #1132, opened 2023-12-29
Early return with null for a ray with no direction

btipling:

### DIFF
--- a/src/math/ray.zig
+++ b/src/math/ray.zig
@@ -76,7 +76,6 @@ pub fn Ray(comptime Vec3P: type) type {
                         @abs(ray.direction.v[1]),
                         @abs(ray.direction.v[2]),
                     });
-                    std.debug.assert(ray.direction.v[kz] != 0.0);
                     if (ray.direction.v[kz] == 0.0) {
                         return null;
                     }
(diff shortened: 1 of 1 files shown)


src/math/ray.zig
```diff
@@ -76,6 +76,10 @@ pub fn Ray(comptime Vec3P: type) type {
                         @abs(ray.direction.v[1]),
                         @abs(ray.direction.v[2]),
                     });
+                    std.debug.assert(ray.direction.v[kz] != 0.0);
```
btipling, line 79: I had a test, but removed it in favor of this assert, which I can't figure out how to have and also have a test that passes, the test I had was:

```zig

test "triangleIntersect_ray_no_direction" {
    const a: math.Vec3 = math.vec3(0, 0, 0);
    const b: math.Vec3 = math.vec3(1, 0, 0);
    const c: math.Vec3 = math.vec3(0, 1, 0);
    const ray: math.Ray = math.Ray{
        .origin = math.vec3(0.1, 0.1, 1),
        .direction = math.vec3(0.0, 0.0, 0.0),
    };

    const result = ray.triangleIntersect(
        &a,
        &b,
        &c,
        true,
    );

    try testing.expect(?math.Ray.Hit, null).eql(result);
}
```

btipling: I had a debug assert but it will remove this check if I do that and result in UB.

slimsag (approved):
Great, thank you!

PR #1182, opened 2024-03-10
module system revamp

### LINKED ISSUE #1169: finish and document our novel/unique event system
## Important

The point of this issue is to share _what I am exploring_ with others interested in Mach because this is a foundational part not yet established. We're exploring this path, and this is the direction we're going in for now.

This issue is to catch others up on where we are going, not an invitation to discuss ad-nauseam the different approaches to ECS (in my experience, there are a _lot_ of people who are very interested in discussing ECS as a code-golf/architecture challenge while not solving real problems with it.)

> meta: the point of that statement is to say that this is the direction we're currently exploring, at this point more discussion of ECS won't help us progress that exploration - we need code improvements and actual testing to further the discussion. It's not to say 'you are not welcome here' but rather to say 'I can't spend a ton of time discussing this at length, we want to play with it and experiment and find out' - it's an apology, not a 'go away' statement 🙂 

## Event system

events (aka message passing) are our (yet-to-be-proven, novel/unique, currently less-than-half-implemented) approach to solving a lot of problems with typical ECS designs. 

Mach modules will support sending events to one another. This is partially implemented today, with global (one->many) and module-to-module (one->one) events.

### Order-of-execution challenges

ECS design patterns often face a challenge of order-of-execution with systems: e.g. you may want physics systems to run before rendering systems as a trivial example.

In other ECS you often describe systems as executing in a specific order via e.g. a sorting integer, or by specifying their dependent systems that should execute first. Kind of like adding all functions to a big `[][]fn`, then executing them in order (using the 2nd list as 'systems that can be run in parallel').

Our solution to this is different, and based on the fact that a message send() will be a synchronization point in the sense that in A->B or A->BC, B and C will be guaranteed to execute after A.

In other words, our approach is a bit like 'call foo() then bar() if you want them to execute in that order' while the traditional approach is a bit more like 'set the priority of foo=1 and bar=2 if you want them to execute in that order, and don't forget about baz=2!'

### Parallelism / multi-threading

Traditionally ECS requires some up-front declarative API design:

* Specifying which order systems should execute in
* Specifying which systems have dependencies on other systems
* Specifying specifically what ECS data will be queried (potentially down to a specific set of entities, components, and whether you will write or just read component data)
* Potentially specifying (or deferring until end-of-frame) creation of new entities.

Traditionally the ECS system scheduler takes all of this information into account when scheduling systems, because if enough information is declared ahead of time then one can work out which systems can execute in parallel without needing e.g. heavy-handed row-level locks on component data for entities to avoid data races.

The downside to this traditional approach is that it requires you to be much more declarative in using the ECS API, which I believe makes it harder to iterate quickly. Additionally, it leads to some 'awkward' situations where one may need to e.g. break up a system that is logically just doing one thing, into two independent systems to fit into the requirements of ECS' declarative patterns.

Again, our solution to this is different:

* The plan is to leverage sending events as an opportunistic point of parallelism. For example, every frame a global `.tick` event is sent to every system that wishes to listen for it, a one->many event, and this would be an opportunity point to execute all systems listening for that event in parallel on different threads.
* As a result, a system in our ECS is not 'a function' but rather 'an event handler' (a function with optional data sent to it)
* As a result, a system in our ECS does not need to declare ahead-of-time what exactly it will query and why. Instead, it can do as it likes - as the programmer intends - and ECS can be the good 'help me structure my code' abstraction that it tries to be.
  * As a consequence, we either (1) have to accept data races if you run two systems in parallel that access the same data - or (2) come up with an alternative system to prevent this without harming performance. We choose (2)
  * Our belief is that with some DB-inspired techniques (think: row locks), additional tooling (to observe event sending behavior, and measuring different parallel execution orders at runtime for contention) - that we can do some runtime-guided-optimization of parallelism points in our scheduler and get the best of both worlds.
  * This is, in many ways, similar to the tradeoffs Zig makes compared to Rust's borrow checker - for example.

### Other use-cases

A keen reader of all the above will have observed something: If all things in our game/app state are derived from _named events and data_ - then surely we could leverage this information for some interesting use cases?

* Observing events as a method of debugging, visualizing and profiling execution flow
* Sending events from another process to control your game/app (e.g. game editor sending events to your game)
* Replaying events to arrive at the same effective game state, if you're careful, such as in e.g. networking code
* Observing changes to entities at whatever level of granularity your prefer (not as 'an observer callback' but as an "I did something" event to be observed)
* Solving the 'new entity creation is deferred until the end of the frame' problem in a different way.
* Solving the traditional (e.g. Unity) mess of 'how do I run something before Update? Before FixedUpdate? Is there a PreUpdate? PrePreUpdate? PrePreFixedUpdate?'
* Running different things at different frequencies - if all things stem from events, one can easily have multiple different loops firing different 'tick' events running at different frequencies, e.g. physics at 240hz and rendering at 60hz, with some synchronization point.
* Modularity in general, because it opens the opportunity for modules to communicate with each-other directly without passing data through entities-and-components from one system to another.

slimsag:
## Summary

Complete revamp of the module and events system.

Closes #1169

### Event handler execution

All events are now asynchronously dispatched, i.e. `.send()` pushes an event onto a queue and someone else pops and executes the function call (previously it was an effectively-inlined function call.)

All (Zig) global and local event handlers are known at comptime, and as a result this bakes down to two single `u32`s describing the module name and event name, respectively, and dispatch is just a large switch statement over the possible global/local event handlers.

The point of this change is to open the opportunity for parallel execution of event handlers across different threads later on: https://github.com/hexops/mach/pull/1182#issuecomment-2015766069 - note that this change does not implement it, only ensuring that we have the means and data structures to do so in the near future.

### Event handler arguments

Event handler arguments will in the future have serialization constraints applied to them, i.e. sending a pointer as an argument to an event handler will be a big no-no.

In general, the idea is that event _arguments_ should be reserved for e.g. simple data, like an editor talking to your game, rather than for passing state between modules. If a module sends state as part of an event, that is now a code smell and discouraged.

### Event handler registration

Previously event handlers were simply any function declared inside of a module, e.g. `foo` (global event handler) and `bar` (local event handler) here:

```
pub fn foo() void {}

pub const local = struct {
    pub fn bar() void {}
};
```

The event handler execution changes require that when we send e.g. a global event, we must know the type of arguments the handler on the other side will expect. In some cases, a module may want to send a global event that nobody is actually handling yet (for example, a global `.tick` event is unhandled when you are just starting out writing your very first module) - and so we need a way to pre-declare what arguments such an event handler _should_ have.

Additionally, parallel deterministic execution of event handlers (called systems in other ECS) require that we have some knowledge basis of certain constraints - [what an event handler will and will not do](https://github.com/hexops/mach/pull/1182#issuecomment-2015766069) in order to determine whether two systems can run in parallel safely and deterministically. We need a way to annotate which constraints an event handler has.

Finally, the previous approach mean that enum names and event handler names would violate Zig name conventions. e.g. an event handler `pub fn updateSprites()` would require `.send(.updateSprites)` to be written rather than the more idiomatic `.send(.update_sprites)`

As a result, we change event handler to a registration model - where `pub const global_events = .{};` declares which global, and `pub const local_events = .{};` declares which local - event handlers there are. e.g. to match the prior code snippet:

```
pub const global_events = .{
    .foo = .{ .handler = foo },
};
pub const local_events = .{
    .bar = .{ .handler = bar },
};

pub fn foo() void {}
pub fn bar() void {}
```

This gives us a place to annotate event handlers with more information (e.g. constraints or documentation strings) in the future.

Because the name is explicitly written different from the function name, one can write the event enum name and handler function name differently to match Zig style guidelines:

```
pub const local_events = .{
    .update_sprites = .{ .handler = updateSprites },
};
```

Finally, it lets one pre-declare what arguments a currently-unhandled event would have by writing the function type rather than providing a pointer to an actual function:

```
pub const local_events = .{
    .update_sprites = .{ .handler = fn (some_arg: u32) void },
};
```

### Component registration

We had a similar issues with component registration:

```
pub const components = struct {
    pub const location = Vec3;
    pub const rotation = Vec3;
    pub const scale = f32;
};
```

1. According ot Zig style guide it should be `pub const Location = Vec3;` .. but that would mean `setComponent(.Location, foo)` which violates the style guide for enums.
2. Although we can write doc comments `/// Location is an entity location` we could not retrieve this information programmatically, e.g. an editor would want to know the description of a component which means the game/app needs to be able to access that information to broadcast it e.g. over the network.
3. It would be weird for `pub const events` and `pub const components` to be written in two different styles.

As a result, we change it to match how events are written:

```
pub const components = struct {
    .location = .{ .type = Vec3 },
    .rotation = .{ .type = Vec3 },
    .scale = .{ .type = f32 },
};
```

One can imagine us adding an optional `.description` string field, or `.default = value` at a later date.

### Module state changes

Previously a module's `.state` field was a copy of the actual module struct's fields. This was intentional to signal that we intended for state to be serializable, and it should not e.g. contain pointers. However, in practice modules often do need pointer state - and so it was always the case that we would need 'some serializable state' and 'some non-serializable state'

As a result, the `.state` field is now _literally_ the module's struct type - rather than just a copy of the fields - which means one can invoke methods of a module's struct via the `.state` field.

Later on, we will add the ability for modules to serialize/deserialize their state as they desire - or signal that parts of it can only be reproduced via the correct sequence of stateful events.

Module state is now also accessed via getters `.state()` or `.stateReadOnly()` - which signals intent of an event handler to read/write state. This information can help us skip replay of some events during record-and-replay in the future.

Additionally, it lets us ensure state has been initialized. `.state()` will panic for example if `foo_mod.init(.{})` was not invoked before. This is less error prone than a default `undefined` initialization, and means that default values now work too.

### Allocator changes

The default/global module allocator, `engine.allocator` and `foo_mod.allocator` for example, has been removed. Modules should use their own allocator if needed.

### ECS code move

`src/ecs` has moved to `src/module`, and things that used to be accessible like e.g. `mach.ecs.Archetype` or `mach.ecs.EntityID` are now top-level, first-class concepts accessible as `mach.Archetype` or `mach.EntityID`.

The motivation of this change is to make the Mach module system a first-class property of Mach, and help clarify some of the differences of our somewhat novel approach to ECS.

## Merge checklist

- [ ] examples no longer abuse event arguments
- [ ] text example allocations no longer leak
- [ ] `// TODO: comptime cleanup` has been resolved everywhere
- [x] `ecs/` namespace makes sense or has been integrated into `module`
- [x] Dispatch loop is an explicit set of events
- [x] `dispatchNoError()` has been removed
- [x] Global allocator (`Modules.allocator`) is gone, allocations strategy for modules is sound
- [x] basic event docs are in place
- [x] Feel good about default module state values
- [x] `zig build run-glyphs` crash on cmd+q exit fixed
- [x] Feel good about global `mach.Module` type / module registration requirements
- [x] Injected args cannot ever have a cyclic type dependency
- [x] `TupleHACK` has been removed
- [x] `pub const events` and `pub const components` use `.name = .{}` syntax
- [x] `ecs/systems.zig` has been removed/integrated into `module.zig`
- [x] `pub const components` and `pub const events` match in how they are written

---


slimsag:
TODO: capture traces at time-of-send
 
image

slimsag:
## Notes

### Requirements for deterministic and parallel execution of event-handlers/system functions

Deterministic execution in parallel requires that we be able to associate the following information with event handler / 'system' functions:

1. Whether it will send events (schedule systems to run), and which?
2. Whether it will read / write entities' components values
3. Whether it will add/remove components
4. Whether it will add/remove entities
5. Whether it will read/write state, and of which modules

The more fine-grained and detailed information we have on these, the more opportunities for parallelism there are in scheduling. The downside is it makes writing systems much more strict and tedious, often requiring people to express these constraints of what a system may do ahead of time. Other ECS handle this as part of writing a system's function signature / parameters.

For example with detail of information, if two systems are known to write to entities' `health_points` component.. some ECS implementations may say that alone is the level of detail you provide - and it implies the scheduler cannot run these two systems at the same time (as it would potentially be a race condition + non-deterministic.)

Some ECS additionally enforce that an ECS run on _sets of entities_ rather than _all entities in the world_ at once, e.g. perhaps your system function is invoked three times each with 1/3rd of the entities in the world - allowing for some parallelism even when two or more systems intend to write to the same e.g. `health_points` component - because as long as they do not do so on the same entities at the same time, then no problem. However, this introduces some restrictions on the programmer (you cannot assume you have a view of the entire world, just a subset of it, within a single execution of your system function) - and that has implications in how people must write more complex systems.

You can also imagine more complex scheduling capabilities depending on more advanced constraints. For example, one system says it only writes to `health_points` if the value is currently set to `0`, and the other says it only writes to `health_points` if the value is currently set to `100` - in this case the scheduler would have sufficient evidence to prove the two do not conflict and could run them in parallel without race conditions. Such scheduling either requires complete static knowledge of what a system does (which would be the best case scenario for an ECS scheduler) - or fairly extensive e.g. constraint annotations on what a system function may do. Many ECS do not reach this level of granularity due to what it would impose on the programmer writing system functions.

PR #1185, opened 2024-03-27
ECS: @memcpy Overlap Memory Bug Fix

Vulfox:
When attempting to remove some items from the entities list, I was receiving some aliasing errors from @memcpy. This happened because there are attempts to copying overlapping regions of memory (same index as the last element in this case) which is not allowed with that builtin function. This fix should alleviate that.

### DIFF
--- a/src/ecs/Archetype.zig
+++ b/src/ecs/Archetype.zig
@@ -195,7 +195,7 @@ pub fn getDynamic(storage: *Archetype, row_index: u32, name: StringTable.Index, 
 
 /// Swap-removes the specified row with the last row in the table.
 pub fn remove(storage: *Archetype, row_index: u32) void {
-    if (storage.len > 1) {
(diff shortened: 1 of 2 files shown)


Vulfox:
I meant to create a separate PR for the invalidated pointer bug, but accidentally scuffed my git commands. If you want me to split them up into 2 different PRs, I can do so.

These are the two bugs I came across while setting up one of my projects. I'm assuming these haven't come up before in other projects cause I am using more than 8 unique archetypes and std.UnmanagedArrayList defaults to having 8 as the minimum capacity. Once I reached 9 elements, the prev_archetype pointer became invalidated as addition memory was required for the append call.

slimsag:
Thanks very much for this! I will pull it into #1182

slimsag:
Merged into #1182 which will be merged soon.

PR #1187, opened 2024-03-30
ecs: add query any

Vulfox:
A basic change-up of the query `all` to be used for the `any` query. Added a test that seems to be on par with the `all` test as is. If a separate test should be created for `any`, let me know. Also if this is not the intended design of `any` with what y'all have in mind, I am also down to adhering to any feedback.

What I really wanted was a query that would allow me to query all entities given a namespace, but figured an any call with all provided component names will do for now and saw it was panic-ing due to no implementation.

### DIFF
--- a/src/ecs/Archetype.zig
+++ b/src/ecs/Archetype.zig
@@ -195,7 +195,7 @@ pub fn getDynamic(storage: *Archetype, row_index: u32, name: StringTable.Index, 
 
 /// Swap-removes the specified row with the last row in the table.
 pub fn remove(storage: *Archetype, row_index: u32) void {
-    if (storage.len > 1) {
+    if (storage.len > 1 and row_index != storage.len - 1) {
         for (storage.columns) |column| {
             const dstStart = column.size * row_index;
             const dst = column.values[dstStart .. dstStart + column.size];
--- a/src/ecs/entities.zig
+++ b/src/ecs/entities.zig
@@ -677,7 +677,25 @@ pub fn ArchetypeIterator(comptime all_components: anytype) type {
                     }
                     return true;
                 },
-                .any => @panic("TODO"),
+                .any => {
+                    for (iter.query.any) |namespace| {
+                        switch (namespace) {
+                            inline else => |components| {
+                                for (components) |component| {
+                                    if (@typeInfo(@TypeOf(component)).Enum.fields.len == 0) continue;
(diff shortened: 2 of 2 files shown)


slimsag:
Thanks for sending this, I am reworking querying quite a bit as part of https://github.com/hexops/mach/pull/1182 so may not merge this as-is due to some conflicts.

What I am very keen to understand is your use-case more specifically, to make sure I cover it. Can you share more details around it? What do the entities look like, and what's the motivation for querying all of them?

Vulfox:
Sure. I am dabbling in making a GUI lib with the ECS system you have here mostly as a learning project for using ECS and what entails making GUIs from scratch. With that said, I am highly certain that whatever pattern I have going on now ain't exactly the most efficient at what I would need it to be. The first step is collecting user input which is just polling and emitting whichever ones I care about and creating an entity under an `Input` namespace corresponding to whatever input was given like so:

https://github.com/Vulfox/guecs/blob/286002525e1726678bf2fbe5ede7d7580d9af2f1/src/Input.zig#L11-L23

To ensure the user input stuff gets cleaned up each loop, I want all `.input` namespaced components gone. Currently it looks like this:

https://github.com/Vulfox/guecs/blob/286002525e1726678bf2fbe5ede7d7580d9af2f1/src/Input.zig#L31-L43

What I thought would have worked for the query was this:

```
var archetypes_iter = engine.entities.query(.{
            .all = &.{ .{ .input = &.{} } },
        });
```

Where I want to select all of the input namespace, this way I don't have to individually be looking for each type of component. However the above seems to select all entities regardless of the namespace. Maybe a bug, but I am assuming it is intended.

I am currently not a fan of polling and emitting user input as I currently have it, but I do like the idea of creating a custom way to create inputs be it for automation or to allow for other input sources that aren't default to mouse and keyboard to create entities that correspond to a given action. This is a naive attempt, and I am currently thinking of ways of how I can reduce the amount of archetype lookups needed to get my input queue cleaned up and that might just need me to reduce the components down to 1 or 2 things which might become a list of some sort.

slimsag:
Okay, this makes a lot of sense and seems like a good use-case to support.

Let me get back to you once I wrap up my work on #1182 to see how we can support this more optimally

slimsag:
Taking this into consideration in a new API design for mach.Core

PR #1194, opened 2024-04-16
math: simplify algorithm finding greatest dimension vector

LiamSwayne:
The `maxDim` function could be further optimized to remove an unnecessary check. The removed check used to observe whether `v[2]` was already greater than `v[0]`, but this is already known because in order for the function to reach that point `v[1]` has to be greater than or equal to `v[0]` and `v[1]` has to not be greater than `v[2]`, so the greatest dimension must be `v[2]` at that point. The removal of this check also removes the need for the final `return 0;` because the else case covers the case in which all three are the same length. This should result in a small performance gain.

The function is still a bit hard to read, which is probably how the unnecessary check snuck in.

### DIFF
--- a/src/math/ray.zig
+++ b/src/math/ray.zig
@@ -47,10 +47,9 @@ pub fn Ray(comptime Vec3P: type) type {
                         }
                     } else if (v[1] > v[2]) {
                         return 1;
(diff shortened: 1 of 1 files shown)


slimsag (approved):
Thanks!

PR #1206, opened 2024-05-11
sysaudio: fix wasapi format selection + fix crash without capture device

Andoryuuta:
This fixes a crash in the `play-opus` example on my Windows 11 machine, wherein no audio format is selected and crashes when trying to read from `device.formats[0]`.
(Stacktrace for reference: https://gist.github.com/Andoryuuta/defa7bef64bfd5d14e913e5dd9b04284)

Additionally, while debugging this, I found that the example would also crash on startup when my microphone was not plugged in. Small drive-by fix for that is included in this MR as well.

This "works on my machine", but I'm not confident I understand the intentions of the previous code enough to say that this is completely correct.


alichraghi (approved):
The actual issue comes from the fact that we temporary removed `i24` format support but the following changes looks good!

slimsag (approved):

PR #1223, opened 2024-07-12
Core: merge old core with ECS Core module

alichraghi:


slimsag:
I am going to blindly merge this with trust for now, it is likely a step in the right direction and it begins moving us away from GLFW which is fantastic.

However, there are several changes in here I don't like and will likely change soon.

PR #1241, opened 2024-08-08
Win32 events and borderless fullscreen

hordurj:
More features added to the win32 platform. Note, the fullscreen mode is currently implemented as borderless full screen. It is not exclusive fullscreen mode which would also change the display mode. 

* Added borderless fullscreen mode.
* Added borderless mode.
* Implemented keyPressed, keyRelease, ... functions
* Added mouse wheel event.
* Added modifiers to keyboard and mouse events.
* Default scaling is stretch in d3d12 backend.
* Display mode borderless and fullscreen behave the same for now.
* The fullscreen implementation will be reviewed later to include
display mode change.

### DIFF
--- a/src/Core.zig
+++ b/src/Core.zig
@@ -130,16 +130,12 @@ pub const InitOptions = struct {
 };
 
 fn init(core: *Mod, entities: *mach.Entities.Mod, options: InitOptions) !void {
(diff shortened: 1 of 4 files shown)


slimsag:
Rebased and merging here: https://github.com/hexops/mach/pull/1255

PR #1247, opened 2024-08-16
Initialize macOS app delegate and set up a Metal view

mjbshaw:
This is the smallest change I could make to get the `run-triangle` demo working on macOS. We can rework this significantly over time but this should at least unblock ongoing development.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -77,6 +77,17 @@ pub fn build(b: *std.Build) !void {
     });
     module.addImport("build-options", build_options.createModule());
 
+    if (target.result.isDarwin()) {
+        if (b.lazyDependency("mach_objc", .{
+            .target = target,
+            .optimize = optimize,
+        })) |dep| {
+            if (want_core or want_sysaudio or want_sysgpu) {
+                module.addImport("objc", dep.module("mach-objc"));
+            }
+        }
+    }
+
     if (want_mach) {
         // Linux gamemode requires libc.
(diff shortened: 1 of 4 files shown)


build.zig
```diff
@@ -77,6 +77,17 @@ pub fn build(b: *std.Build) !void {
     });
     module.addImport("build-options", build_options.createModule());
 
+    if (target.result.isDarwin()) {
+        if (b.lazyDependency("mach_objc", .{
```
slimsag, line 81: note: I will undo this specific change, but wanted to let you know why:

you might not have encountered this yet, but the way `b.lazyDependency` works is that if it is invoked, that dependency will be downloaded - so the `if (want_core or want_sysaudio or want_sysgpu)` being inside the if statement rather than outside means it is always fetched - even if it wouldn't be used.

finally, I prefer to have dependencies like this inline under each `want_core`, etc. case to be a bit more explicit about who depends on what - even if it's a little bit of duplicative code - because it tends to be more readable long-term

slimsag (approved):
super helpful, thank you!

build.zig
mjbshaw: Thanks for the explanation!

Regarding having the deps inline to be more explicit about what depends on what: that makes sense, but the main reason I changed this bit wasn't because of the duplicative code itself, but rather the duplicative work executed. But reviewing Zig's `std.Build.dependencyInner()`'s source it looks like it has duplicate detection so I suppose it's not too concerning.

PR #1250, opened 2024-08-21
Update to the latest mach-objc

mjbshaw:
This eliminates the dual namespacing, runtime initialization, and dependency on `usingnamespace`


mjbshaw:
Whoops, the first commit had some bad `as()` casts that I missed in testing only `run-triangle`. Fixed in the latest force push.

slimsag (approved):
fantastic work!

Issue #1128, opened 2023-12-25
sysgpu: Crash on run on Windows


foxnne:
When using sysgpu, there is an immediate crash when running on Windows.

```
 *  Executing task: zig build run -Duse_dusk=true 

steps [27/30] zig build-exe pixi Debug native... LLD Link... thread 6068 panic: api error
C:\Users\foxnn\AppData\Local\zig\p\1220d0bf732d8291b8629ce7e115efb6adef84dcc1218317ee0a89a6f1663153f2a9\src\d3d12.zig:89:13: 0x7ff6fb24eae0 in init (pixi.exe.obj)
            return error.CreateDXGIFactoryFailed;
            ^
C:\Users\foxnn\AppData\Local\zig\p\122096b39d9266d275b780df4d63b1afe6e9a3726f19766baf74b34af3f092532cd3\src\platform\native\Core.zig:202:40: 0x7ff6fb255a8d in init (pixi.exe.obj)
    const instance = gpu.createInstance(null) orelse {
                                       ^
C:\Users\foxnn\AppData\Local\zig\p\122096b39d9266d275b780df4d63b1afe6e9a3726f19766baf74b34af3f092532cd3\src\main.zig:155:27: 0x7ff6fb2598b0 in init (pixi.exe.obj)
    try platform.Core.init(
                          ^
C:\Users\foxnn\dev\proj\pixi\src\pixi.zig:134:18: 0x7ff6fb25f87e in init (pixi.exe.obj)
    try core.init(.{
                 ^
C:\Users\foxnn\AppData\Local\zig\p\122096b39d9266d275b780df4d63b1afe6e9a3726f19766baf74b34af3f092532cd3\src\platform\native\main.zig:36:17: 0x7ff6fb265e0d in main (pixi.exe.obj)
    try app.init();
                ^
C:\Users\foxnn\.zvm\master\lib\std\start.zig:482:101: 0x7ff6fb266a79 in main (pixi.exe.obj)
    return @call(.always_inline, callMainWithArgs, .{ @as(usize, @intCast(c_argc)), @as([*][*:0]u8, @ptrCast(c_argv)), envp });
                                                                                                    ^
C:\Users\foxnn\.zvm\master\lib\libc\mingw\crt\crtexe.c:321:0: 0x7ff6fb62b385 in __tmainCRTStartup (crt2.obj)
    mainret = main (argc, argv, envp);

[...]

PR #1251, opened 2024-08-22
Show warning instead of panic when debug layer cant be created #1128

hordurj:

### DIFF
--- a/src/sysgpu/d3d12.zig
+++ b/src/sysgpu/d3d12.zig
@@ -84,7 +84,19 @@ pub const Instance = struct {
             &c.IID_IDXGIFactory4,
             @ptrCast(&dxgi_factory),
         );
-        if (hr != c.S_OK) {
+        if (hr == c.DXGI_ERROR_INVALID_CALL) {
+            const hr_prev = hr;
+            hr = c.CreateDXGIFactory2(
+                0,
+                &c.IID_IDXGIFactory4,
+                @ptrCast(&dxgi_factory));
+            if (hr == c.S_OK) {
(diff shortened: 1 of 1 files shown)


src/sysgpu/d3d12.zig
```diff
@@ -84,11 +84,28 @@ pub const Instance = struct {
...
+                &c.IID_IDXGIFactory4,
+                @ptrCast(&dxgi_factory));
+            if (hr == c.S_OK) {
+                std.debug.print(
+                    \\Failed to enable debug layer ({x}):
+                         \\ For Windows 10, to create a device that supports the debug layer, enable the "Graphics Tools" optional
+                         \\ feature. On Windows 10 22H2, go to the Settings panel, under System. On older versions of Windows 10,
+                         \\ go to the Settings panel, under Apps > Apps & features. Go to Optional Features > Add a feature, and
+                         \\ then look for "Graphics Tools"
+                         \\
```
slimsag, line 100: 1. Can you use `log.info` instead of `std.debug.print`? That lets others silence this warning if they want, etc. and is more consistent with what we do elsewhere in this file.

2. Could you also add a link to the Microsoft article here at the end? In case this info becomes out of data it'd be useful probably.

slimsag: Actually, thinking about it a bit more, maybe let's go with a shorter and simpler message using `log.info`:

> note: D3D12 debug layers disabled (couldn't enable, error: {x}), see https://machengine.org/about/faq/#how-to-enable-direct3d-debug-layers

And then I can add this info to the FAQ on the website.

slimsag (approved):

PR #1256, opened 2024-08-25
core: remove linux support

slimsag:
The `mach.Core` API needs major design changes, and every backend that it supports today must be updated to account for those API design changes. Unless someone is actively maintaining and improving the state of a given backend, it slows down our ability to make those critical changes.

Unfortunately, the backends for X11 and Wayland today are half-baked, nobody is actively maintaining on or contributing to them, and the Linux CI tests have been broken for over a month as a result which harms overall stability of Mach.

As a result, this PR removes Linux support from `mach.Core`


slimsag:
Excellent, this has already surfaced regressions in cross-compilation from Linux -> Windows.

PR #1264, opened 2024-09-07
Fix sysgpu_backend build option 

Tatskaari:
The sysgpu backend build option was treating the build option as the wrong enum type. This type is defined in `build.zig` and is exposed through the generated `build_options` struct. This is really nasty, but I think this is the correct way to do this. 

The generated code for `@import("build_options")` for reference:
```
pub const want_mach: bool = true;
pub const want_core: bool = true;
pub const want_sysaudio: bool = true;
pub const want_sysgpu: bool = true;
pub const @"build.SysgpuBackend" = enum (u3) {
    default = 0,
    webgpu = 1,
    d3d12 = 2,
    metal = 3,
    vulkan = 4,
    opengl = 5,
};
pub const sysgpu_backend: @"build.SysgpuBackend" = .opengl;
pub const @"build.Platform" = enum (u3) {
    wasm = 0,
    linux = 1,
    windows = 2,
    darwin = 3,
    null = 4,
};
pub const core_platform: @"build.Platform" = .linux;
```

### DIFF
--- a/src/sysgpu/main.zig
+++ b/src/sysgpu/main.zig
@@ -5,7 +5,20 @@ pub const sysgpu = @import("sysgpu/main.zig");
 pub const shader = @import("shader.zig");
 const utils = @import("utils.zig");
 
-const backend_type: sysgpu.BackendType = backendFromBuildOption(build_options.sysgpu_backend);
+const backend_type: sysgpu.BackendType = switch (build_options.sysgpu_backend) {
+    .webgpu => .webgpu,
+    .d3d12 => .d3d12,
+    .metal => .metal,
+    .vulkan => .vulkan,
+    .opengl => .opengl,
+    .default => switch (builtin.target.os.tag) {
+        .linux => .vulkan,
+        .macos, .ios => .metal,
+        .windows => .d3d12,
+        else => @compileError("unsupported platform"),
+    },
+};
+
 const impl = switch (backend_type) {
     .d3d12 => @import("d3d12.zig"),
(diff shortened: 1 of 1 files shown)


src/sysgpu/main.zig
```diff
@@ -23,6 +17,24 @@ const impl = switch (backend_type) {
 var inited = false;
 var allocator: std.mem.Allocator = undefined;
 
+fn backendFromBuildOption(comptime opt: build_options.@"build.SysgpuBackend") sysgpu.BackendType {
```
slimsag, line 20: This being named `build_options.@"build.SysgpuBackend"` looks weird and wrong to me.

We have other options in `build_options` that don't have to be written that way, for example here you can see we have `build_options.want_sysaudio` as a boolean: https://github.com/hexops/mach/blob/main/src/main.zig#L18

I think we do something similar for `core.PlatformType`: https://github.com/hexops/mach/blob/main/build.zig#L48

Can you look into making it work similar to that?

slimsag: I think the solution might be writing something like:

```
const backend_type: sysgpu.BackendType = switch(build_options.sysgpu_backend) {
    // ... all the cases ...
};
```

And then don't have a `backendFromBuildOption` function at all.

Tatskaari: Ah yeah that would be nicer. I've updated the PR :)

slimsag (approved):
Looks great, thank you for fixing this!

PR #1266, opened 2024-09-07
Implement `MatShared.eql()` and `MatShared.eqlApprox()` methods

joshua-holmes:
Implement `.eql()` and `.eqlApprox()` methods for matrices. The previous implementation only supported a single method for comparing matrices, but I thought it might be useful to include two methods, one that uses elipson tolerance and another that uses user-defined tolerance, just like how `Vector`s currently handle it.

I felt it was redundant to include tests for 2x2 and/or 3x3, but let me know if you want me to add those.

### DIFF
--- a/src/math/mat.zig
+++ b/src/math/mat.zig
@@ -483,6 +483,8 @@ pub fn Mat4x4(
 
         pub const mul = Shared.mul;
         pub const mulVec = Shared.mulVec;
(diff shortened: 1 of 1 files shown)


slimsag (approved):
Excellent, thank you!

joshua-holmes:
fyi, fixed bad wording in my code comments

Issue #1265, opened 2024-09-07
Core: build succeeds on linux but fails at runtime


Tatskaari:
Linux support has been removed in this commit: 
https://github.com/hexops/mach/commit/17450fde65289da7ac01ba174c26687fc9691182 

It's not clear when building on linux that core is no longer supported. Should we fail fast here?

joshua-holmes:
Since you posted this, Linux support has been added again on Wayland. I am working on X11 support soon as well :)

joshua-holmes:
@Tatskaari after [this PR](https://github.com/hexops/mach/pull/1282) is pulled in, X11 will be supported as well, though some features are missing. However there is a log message that is sent that lists the missing features.

PR #1268, opened 2024-09-13
core: fix and enable wayland

joshua-holmes:


src/core/linux/Wayland.zig
```diff
@@ -21,53 +21,92 @@ pub const c = @cImport({
...
+    return @call(.always_tail, libwaylandclient.wl_proxy_destroy, .{proxy});
+}
+
 state: *Core,
+core: *Core,
 title: [:0]const u8,
-size: Core.Size,
-surface_descriptor: gpu.Surface.Descriptor,
-configured: bool,
-decoration: *c.zxdg_toplevel_decoration_v1,
```
joshua-holmes, line 29: didn't need to save this in state, because it is being used in the same scope it is being set

src/core/linux/Wayland.zig
```diff
@@ -21,53 +21,92 @@ pub const c = @cImport({
...
+size: *Core.Size,
+surface_descriptor: gpu.Surface.DescriptorFromWaylandSurface,
+configured: bool = false,
 
 display: *c.wl_display,
-surface: *c.wl_surface,
-registry: *c.struct_wl_registry,
```
joshua-holmes, line 33: didn't need to save this in state, because it is being used in the same scope it is being set

src/core/linux/Wayland.zig
```diff
@@ -21,53 +21,92 @@ pub const c = @cImport({
...
-xdg_surface: *c.xdg_surface,
```
joshua-holmes, line 34: didn't need to save this in state, because it is being used in the same scope it is being set

src/core/linux/Wayland.zig
```diff
@@ -21,53 +21,92 @@ pub const c = @cImport({
...
-toplevel: *c.xdg_toplevel,
```
joshua-holmes, line 35: `decoration`, `registry`, `xdg_surface` and `toplevel` were all properties that were only being used once, and it was in the same scope they were being set, so no sense in saving them in state

src/Core.zig
```diff
@@ -183,14 +183,15 @@ fn init(core: *Mod, entities: *mach.Entities.Mod) !void {
...
 
+    const platform: Platform = undefined;
     core.init(.{
         .allocator = allocator,
         .main_window = main_window,
         .events = events,
         .input_state = .{},
 
         // TODO: remove undefined initialization (disgusting!)
-        .platform = undefined,
```
joshua-holmes, line 193: Had to do this because Zig was unsure of the type of `platform` during compile time, and it needed to know that before passing it into `try Platform.init()` [here](https://github.com/hexops/mach/pull/1268/files#diff-5d71102262b2ae72ea5d47ecbde60a14e74dfea97b9351dd2610021f39b88774L209)

build.zig
```diff
@@ -102,6 +102,32 @@ pub fn build(b: *std.Build) !void {
...
     }
     if (want_core) {
+        if (Platform.fromTarget(target.result) == .linux) {
+            const lib = b.addStaticLibrary(.{
+                .name = "core-wayland",
+                .target = target,
+                .optimize = optimize,
+            });
+            lib.addCSourceFile(.{
+                .file = b.path("src/core/linux/wayland.c"),
```
slimsag, line 112: This is only needed because translate-c can't handle some of the types in the header, right?

slimsag: Filed https://github.com/hexops/mach/issues/1269 as a follow-up issue

slimsag (approved):

build.zig
joshua-holmes: Pretty sure that's the issue, yeah.

joshua-holmes:
Resolves #1265

joshua-holmes:
Closes #1265

PR #1282, opened 2024-10-10
core: enable x11, employ `MACH_BACKEND` env var, minor fixes/tweaks

joshua-holmes:
This PR makes the following changes:
- enable x11
- use `MACH_BACKEND` env var instead of `MACH_CORE_BACKEND` for selecting linux display server
- fix minor window size discrepancy bug in wayland
- add logging for both x11 and wayland that explain that not all features are supported yet


slimsag (approved):
Nice work as usual :)

Issue #1290, opened 2024-10-21
core: X11's update sometimes fails while polling for events


RonaldZielaznicki:
```
[xcb] Unknown sequence number while processing queue
[xcb] You called XInitThreads, this is not your fault
[xcb] Aborting, sorry about that.
core-triangle: ../../src/xcb_io.c:278: poll_for_event: Assertion `!xcb_xlib_threads_sequence_lost' failed.
run-triangle
└─ run core-triangle failure
```

It's pretty random when this happens, but I did isolate it to the `Update` method. If I don't touch the keyboard, I can't reproduce the error.

But, if I scroll the mouse, move windows; I can *eventually* get the above error to occur.

This error will also occur during deinit. But, is hidden by a double free error that is fixed in https://github.com/hexops/mach/pull/1289.

PR #1291, opened 2024-10-21
Call XInitThreads first; XCloseDisplay cleanup

RonaldZielaznicki:

Closes https://github.com/hexops/mach/issues/1290

## Context
yppy provided some excellent advise on discord:
> XInitThreads should be the first xlib function called (though im not sure i would believe in xlib multithreading...)

And after I mentioned an error code on exit, they also advised:
> error 3 is badwindow
> probably something is using it after XDestroyWindow

The discussion after lead to the realization that `platform.deinit` is called *before* the GPU backend is released. This caused a race condition where the GPU backend might interact with the window before it is released.

## Changes

* change: `libx11.XInitThreads` is now called first when initializing X11.
* remove: X11's custom `errorHandler` in favor of the default handler's more verbose description.
* change: `platform.deinit` is now called *after* the GPU backend has been released.

## Some additional detail

To show why I removed the custom X11 error handler.

### Custom Error Message
```
err(mach): X11: error code 3
```

### Default Error Message
```
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  148 ()
  Minor opcode of failed request:  1
  Resource id in failed request:  0x5400002
  Serial number of failed request:  2035
  Current serial number in output stream:  2036
```

### DIFF
--- a/src/core/linux/X11.zig
+++ b/src/core/linux/X11.zig
@@ -86,6 +86,7 @@ pub fn init(
 ) !X11 {
     // TODO(core): return errors.NotSupported if not supported
     const libx11 = try LibX11.load();
+    _ = libx11.XInitThreads();
     const libgl: ?LibGL = LibGL.load() catch |err| switch (err) {
         error.LibraryNotFound => null,
         else => return err,
@@ -185,7 +186,6 @@ pub fn init(
         .libxkbcommon = try LibXkbCommon.load(),
     };
     _ = libx11.XSetErrorHandler(errorHandler);
-    _ = libx11.XInitThreads();
     _ = libx11.XrmInitialize();
     defer _ = libx11.XFreeColormap(display, colormap);
     for (0..2) |i| {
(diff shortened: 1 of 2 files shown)


src/core/linux/X11.zig
```diff
@@ -86,6 +86,11 @@ pub fn init(
 ) !X11 {
     // TODO(core): return errors.NotSupported if not supported
     const libx11 = try LibX11.load();
+
+    // Xlibx11.XInitThreads must be called first
+    //
+    // if not, 'Unknown sequence number while processing queue' errors occur.
+    _ = libx11.XInitThreads();
```
slimsag, line 93: I'd like us to remove this from the codebase ASAP, since we don't have any multi-threading yet this _should_ be doing nothing :) so let's get it out and debug any issues that arise from removing it

slimsag (approved):

PR #1293, opened 2024-10-25
core: wayland: implement keyboard input & window resizing

joshua-holmes:
Changes:
- Implements `Wayland.update()` function based on GLFW implementation
    - Keyboard input now works
    - Window resizing now works
- Fixes bug caused by passing stack reference into xdg and wayland listeners


slimsag (approved):
Looks like a solid improvement, nice work

PR #1297, opened 2024-10-31
core: x11: clarify X11 multi-threading

RonaldZielaznicki:

### DIFF
--- a/src/core/linux/X11.zig
+++ b/src/core/linux/X11.zig
@@ -84,9 +84,11 @@ pub fn init(
     // TODO(core): return errors.NotSupported if not supported
     const libx11 = try LibX11.load();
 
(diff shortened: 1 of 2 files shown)


src/core/linux/X11.zig
```diff
@@ -65,9 +65,6 @@ hidden_cursor: c.Cursor,
 // Mutable fields only used by main thread
 cursors: [@typeInfo(CursorShape).@"enum".fields.len]?c.Cursor,
 
-// Input state; written from main thread; read from any
-input_mu: std.Thread.RwLock = .{},
```
slimsag, line 69: I think that the removal of this mutex is probably wrong; however I expect the external API for input handling to change greatly so totally fine for now.

slimsag (approved):

PR #1307, opened 2024-11-27
obj: Allow tracking of individual struct fields for changes

foxnne:

### DIFF
--- a/src/module.zig
+++ b/src/module.zig
@@ -232,7 +232,7 @@ pub fn Objects(options: ObjectsOptions, comptime T: type) type {
         ///
         /// Unlike set(), this method does not respect any mach.Objects tracking
         /// options, so changes made to an object through this method will not be tracked.
-        pub fn setRaw(objs: *@This(), id: ObjectID, comptime field_name: std.meta.FieldEnum(T), value: anytype) void {
+        pub fn setRaw(objs: *@This(), id: ObjectID, comptime field_name: std.meta.FieldEnum(T), value: std.meta.FieldType(T, field_name)) void {
             const data = &objs.internal.data;
             const unpacked = objs.validateAndUnpack(id, "setRaw");
 
@@ -246,7 +246,7 @@ pub fn Objects(options: ObjectsOptions, comptime T: type) type {
         ///
         /// Unlike setAllRaw, this method respects mach.Objects tracking
         /// and changes made to an object through this method will be tracked.
-        pub fn set(objs: *@This(), id: ObjectID, comptime field_name: std.meta.FieldEnum(T), value: anytype) void {
+        pub fn set(objs: *@This(), id: ObjectID, comptime field_name: std.meta.FieldEnum(T), value: std.meta.FieldType(T, field_name)) void {
             const data = &objs.internal.data;
             const unpacked = objs.validateAndUnpack(id, "set");
 
@@ -262,7 +262,7 @@ pub fn Objects(options: ObjectsOptions, comptime T: type) type {
         }
 
         /// Get a single field.
-        pub fn get(objs: *@This(), id: ObjectID, field_name: anytype) ?std.meta.FieldType(T, field_name) {
+        pub fn get(objs: *@This(), id: ObjectID, comptime field_name: std.meta.FieldEnum(T)) ?std.meta.FieldType(T, field_name) {
             const data = &objs.internal.data;
 
             const unpacked = objs.validateAndUnpack(id, "get");
--- a/src/module.zig
+++ b/src/module.zig
@@ -25,7 +25,13 @@ const PackedObjectTypeID = packed struct(u16) {
     object_name_id: u6,
 };
 
-pub fn Objects(comptime T: type) type {
+pub const ObjectsOptions = struct {
+    /// If this option is true, the internal field `update` will
+    /// contain a list of bitsets representing each object field
+    track_fields: bool = false,
+};
+
+pub fn Objects(comptime T: type, options: ObjectsOptions) type {
     return struct {
         internal: struct {
             allocator: std.mem.Allocator,
@@ -59,6 +65,10 @@ pub fn Objects(comptime T: type) type {
 
             /// Global pointer to object relations graph
             graph: *Graph,
+
+            /// This is mach's way of determining which fields are updated and
+            /// need to trigger some other logic
+            updated: std.ArrayListUnmanaged(std.bit_set.DynamicBitSetUnmanaged) = .{},
         },
 
         pub const IsMachObjects = void;
@@ -149,6 +159,7 @@ pub fn Objects(comptime T: type) type {
             const dead = &objs.internal.dead;
             const generation = &objs.internal.generation;
             const recycling_bin = &objs.internal.recycling_bin;
+            const _updated = &objs.internal.updated;
 
             // The recycling bin should always be big enough, but we check at this point if 10% of
             // all objects have been thrown on the floor. If they have, we find them and grow the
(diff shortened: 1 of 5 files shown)


src/module.zig
```diff
@@ -176,28 +187,112 @@ pub fn Objects(comptime T: type) type {
...
+
+        /// This function will set a single field of an object.
+        ///
+        /// Note that this function also sets a bit set value
+        /// for each object field, informing mach that the field has
+        /// been updated and can be passed to internals that need it.
+        ///
+        /// Example: core.windows.set(id, .width, value) will trigger the Platform
+        /// to update the window's size.
+        pub fn set(objs: *@This(), id: ObjectID, comptime field_name: anytype, value: anytype) void {
```
slimsag, line 266:
```suggestion
        pub fn set(objs: *@This(), id: ObjectID, comptime field_name: std.meta.FieldEnum(T), value: anytype) void {
```

We can use FieldEnum here, which will return an enum for each field in `T`, guaranteeing that what we got is actually a field of the struct (and removing the need for the `hand-written `@compilerError` type check below)

src/module.zig
```diff
@@ -59,6 +65,10 @@ pub fn Objects(comptime T: type) type {
 
             /// Global pointer to object relations graph
             graph: *Graph,
+
+            /// This is mach's way of determining which fields are updated and
+            /// need to trigger some other logic
+            updated: std.ArrayListUnmanaged(std.bit_set.DynamicBitSetUnmanaged) = .{},
```
slimsag, line 71:
```
/// A bitset per-field used to track field changes. Only used if options.track_fields == true.
```

src/module.zig
```diff
@@ -149,6 +159,7 @@ pub fn Objects(comptime T: type) type {
             const dead = &objs.internal.dead;
             const generation = &objs.internal.generation;
             const recycling_bin = &objs.internal.recycling_bin;
+            const _updated = &objs.internal.updated;
```
slimsag, line 162: nit: prefer avoiding underscores, we can rename one to `updated_fields` instead?

src/module.zig
```diff
@@ -176,28 +187,112 @@ pub fn Objects(comptime T: type) type {
...
         }
 
-        pub fn set(objs: *@This(), id: ObjectID, value: T) void {
+        /// This function will set all fields of an object.
+        ///
+        /// Note that this function has no tracking, and will not trigger
+        /// any internal mach functions.
+        ///
+        /// Example: core.windows.setAllRaw(id, value) will NOT trigger the Platform
+        /// to update the window's size, etc.
```
slimsag, line 213: I'd like to avoid the mach.Core-specific example in the docs here. Also, I don't like that we say `will not trigger any internal mach functions` because that makes the module system sound way more magical than it is.

Instead, we can do something like:

```
/// Sets all fields of the given object to the given value.
///
/// Unlike setAll(), this method does not respect any mach.Objects tracking
/// options, so changes made to an object through this method will not be tracked.
```

src/module.zig
```diff
@@ -176,28 +187,112 @@ pub fn Objects(comptime T: type) type {
...
 
-        pub fn get(objs: *@This(), id: ObjectID) ?T {
+        /// This function will set a single field of an object.
+        ///
+        /// Note that this function has no tracking, and will not trigger
+        /// any internal mach functions.
+        ///
+        /// Example: core.windows.setRaw(id, .width, value) will NOT trigger the Platform
+        /// to update the window's size.
+        pub fn setRaw(objs: *@This(), id: ObjectID, comptime field_name: anytype, value: anytype) void {
```
slimsag, line 246: Same thing here w.r.t. comments

src/module.zig
slimsag: Same thing here w.r.t. comments

src/module.zig
```diff
@@ -176,28 +187,112 @@ pub fn Objects(comptime T: type) type {
...
+        /// Example: core.windows.setAll(id, value) can trigger the Platform
+        /// to update the window's size, etc.
+        pub fn setAll(objs: *@This(), id: ObjectID, value: T) void {
+            const data = &objs.internal.data;
+
+            const unpacked = objs.validateAndUnpack(id, "setAll");
             data.set(unpacked.index, value);
+
+            if (options.track_fields)
+                objs.internal.updated.items[unpacked.index].setAll();
```
slimsag, line 236: nit: prefer keeping these on a single line / not wrapping to the next line

src/module.zig
slimsag: We could make this field nullable and have it be null if track_fields == false? Not sure if that's better memory wise but maybe worth considering.

src/module.zig
slimsag: `value: anytype` here and in other places can be `value: T` I think for better type safety?

src/module.zig
```diff
@@ -25,7 +25,13 @@ const PackedObjectTypeID = packed struct(u16) {
...
 };
 
-pub fn Objects(comptime T: type) type {
+pub const ObjectsOptions = struct {
+    /// If this option is true, the internal field `update` will
+    /// contain a list of bitsets representing each object field
+    track_fields: bool = false,
+};
+
+pub fn Objects(comptime T: type, options: ObjectsOptions) type {
```
slimsag, line 34: Since `options` will always be shorter than `T` I think when writing it, let's put options first?

slimsag: I guess we could expose `mach.Objects(T)` and `mach.ObjectsWithOptions(T, options)`

Might be better.. hmm

src/module.zig
foxnne: Done

src/module.zig
foxnne: The value here is the field value, not the full object value though?

src/module.zig
```diff
@@ -176,28 +187,94 @@ pub fn Objects(comptime T: type) type {
             try dead.resize(allocator, data.capacity, true);
             try generation.ensureUnusedCapacity(allocator, 1);
 
+            if (objs.internal.updated) |*updated_fields| {
+                try updated_fields.ensureUnusedCapacity(allocator, 1);
+                updated_fields.appendAssumeCapacity(try std.bit_set.DynamicBitSetUnmanaged.initEmpty(allocator, @typeInfo(T).@"struct".fields.len));
```
slimsag, line 192: So, right now we have a single `std.bit_set.DynamicBitSetUnmanaged` per object. An interesting idea might be to use a single one for all objects.

The way this would work is we make the bitset `num_objects * @typeInfo(T).@"struct".fields.len`, then when we set bits we do `objs.internal.updated.set((object_index * @typeInfo(T).@"struct".fields.len) + field_index)`

This might be better in terms of memory efficiency, because:

```
//! DynamicBitSet:
//!   A bit set with runtime-known size, backed by an allocated slice
//!   of usize.
```

versus with the array-of-bitsets we store the `DynamicBitSetUnmanaged` (two `usize` values per bitset, plus whatever is needed for storage.)

More importantly, it would reduce the number of small allocations needed

slimsag: That said - I won't block this PR on this :).

slimsag (approved):

src/module.zig
```diff
@@ -239,6 +316,22 @@ pub fn Objects(comptime T: type) type {
             return unpacked;
         }
 
+        /// Returns true if the field has an `updated` bit set in internal, after setting the bit
+        /// back to false.
```
slimsag, line 320: It doesn't make sense to mention `internal` in documentation comments for the public API; internal should be.. well.. an internal implementation detail the user of the API shouldn't need to think or worry about :)

PR #1309, opened 2024-11-30
obj: Move `Platform` and `InitOptions` fields into `core.windows`

foxnne:
This PR is stacked on top of #1308, so review that one first.

At the moment, this PR breaks a lot and only triangle example on macOS is building. The effort here is to move more of Core to use the object system, first of these efforts is to treat windows as a list of objects so all window-dependent fields have been moved to `core.windows` and `Platform.tick` just checks for field changes and responds accordingly.

### DIFF
--- a/examples/core-triangle/App.zig
+++ b/examples/core-triangle/App.zig
@@ -1,3 +1,4 @@
+const std = @import("std");
 const mach = @import("mach");
 const gpu = mach.gpu;
 
@@ -24,43 +25,49 @@ pub fn init(
     core.on_tick = app_mod.id.tick;
     core.on_exit = app_mod.id.deinit;
 
-    // Create our shader module
-    const shader_module = core.device.createShaderModuleWGSL("shader.wgsl", @embedFile("shader.wgsl"));
-    defer shader_module.release();
+    if (core.windows.getAll(core.main_window)) |cw| {
+        const core_window = cw;
+        if (core_window.device) |device| {
 
-    // Blend state describes how rendered colors get blended
-    const blend = gpu.BlendState{};
+            // Create our shader module
+            const shader_module = device.createShaderModuleWGSL("shader.wgsl", @embedFile("shader.wgsl"));
+            defer shader_module.release();
 
-    // Color target describes e.g. the pixel format of the window we are rendering to.
-    const color_target = gpu.ColorTargetState{
-        .format = core.windows.get(core.main_window, .framebuffer_format).?,
-        .blend = &blend,
-    };
+            // Blend state describes how rendered colors get blended
+            const blend = gpu.BlendState{};
 
-    // Fragment state describes which shader and entrypoint to use for rendering fragments.
-    const fragment = gpu.FragmentState.init(.{
-        .module = shader_module,
-        .entry_point = "frag_main",
-        .targets = &.{color_target},
-    });
+            // Color target describes e.g. the pixel format of the window we are rendering to.
+            const color_target = gpu.ColorTargetState{
+                .format = core.windows.get(core.main_window, .framebuffer_format).?,
+                .blend = &blend,
+            };
 
-    // Create our render pipeline that will ultimately get pixels onto the screen.
-    const label = @tagName(mach_module) ++ ".init";
-    const pipeline_descriptor = gpu.RenderPipeline.Descriptor{
-        .label = label,
-        .fragment = &fragment,
-        .vertex = gpu.VertexState{
-            .module = shader_module,
-            .entry_point = "vertex_main",
-        },
-    };
-    const pipeline = core.device.createRenderPipeline(&pipeline_descriptor);
+            // Fragment state describes which shader and entrypoint to use for rendering fragments.
+            const fragment = gpu.FragmentState.init(.{
+                .module = shader_module,
+                .entry_point = "frag_main",
+                .targets = &.{color_target},
+            });
 
-    // Store our render pipeline in our module's state, so we can access it later on.
-    app.* = .{
-        .title_timer = try mach.time.Timer.start(),
-        .pipeline = pipeline,
-    };
+            // Create our render pipeline that will ultimately get pixels onto the screen.
+            const label = @tagName(mach_module) ++ ".init";
+            const pipeline_descriptor = gpu.RenderPipeline.Descriptor{
+                .label = label,
+                .fragment = &fragment,
+                .vertex = gpu.VertexState{
+                    .module = shader_module,
+                    .entry_point = "vertex_main",
+                },
+            };
+            const pipeline = device.createRenderPipeline(&pipeline_descriptor);
+
+            // Store our render pipeline in our module's state, so we can access it later on.
+            app.* = .{
+                .title_timer = try mach.time.Timer.start(),
+                .pipeline = pipeline,
+            };
+        }
+    }
 
     // TODO(object): window-title
     // try updateWindowTitle(core);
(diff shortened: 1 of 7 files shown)


examples/core-triangle/App.zig
```diff
@@ -24,43 +25,49 @@ pub fn init(
...
-    const pipeline = core.device.createRenderPipeline(&pipeline_descriptor);
-
-    // Store our render pipeline in our module's state, so we can access it later on.
-    app.* = .{
-        .title_timer = try mach.time.Timer.start(),
-        .pipeline = pipeline,
-    };
+    if (core.windows.getAll(core.main_window)) |cw| {
+        const core_window = cw;
+        if (core_window.device) |device| {
```
slimsag, line 30: presumably both these if statements, if the logic inside of them doesn't run, the app wouldn't work. So I think we should replace them with non-null assertions:

```
const core_window = core.windows.getAll(core.main_window).?;
const device = core_window.device.?;
```

examples/core-triangle/App.zig
```diff
@@ -69,53 +76,83 @@ pub fn init(
...
+                    .left => {
+                        var w = core.windows.getAll(core.main_window).?;
+                        w.width = w.width - 10;
+                        core.windows.setAll(core.main_window, w);
+                    },
+                    .up => {
+                        if (core.windows.get(core.main_window, .height)) |height| core.windows.set(core.main_window, .height, height + 10);
+                    },
+                    .down => {
+                        if (core.windows.get(core.main_window, .height)) |height| core.windows.set(core.main_window, .height, height - 10);
```
slimsag, line 95: same with these; can use a `.?` non-null assertion?

slimsag: Wait a minute.. I think I typo'd the `get` function signature. I don't think it _can_ return null at all lol.

src/core/Darwin.zig
```diff
@@ -76,193 +59,299 @@ pub fn run(comptime on_each_update_fn: anytype, args_tuple: std.meta.ArgsTuple(@
...
+
+            if (core.windows.get(window_id, .device).? == null) {
+                // Window has not been set up yet, we need to init the window
+                //try initWindow(core, window_id);
+                //try core.initWindow(window_id);
+
+                try initWindow(core, window_id);
+                try core.initWindow(window_id);
+            }
+            if (native.window) |native_window| {
```
slimsag, line 77: Ohhh, you know what we should do: let's use the `native` field as the optional value indicating 'the window was initialized'. Then the `.device` field can be `undefined` prior to that rather than nullable

src/core/Darwin.zig
```diff
@@ -76,193 +59,299 @@ pub fn run(comptime on_each_update_fn: anytype, args_tuple: std.meta.ArgsTuple(@
...
     core: *Core,
-    options: InitOptions,
+    window_id: mach.ObjectID,
 ) !void {
-    var surface_descriptor = gpu.Surface.Descriptor{};
-
-    // TODO: support UIKit.
-    var window_opt: ?*objc.app_kit.Window = null;
-    if (!options.headless) {
+    if (core.windows.getAll(window_id)) |cw| {
```
slimsag, line 94: Oh! I also think that `getAll` cannot return a `null` value ever, we should fix that too

src/Core.zig
```diff
@@ -40,25 +40,61 @@ windows: mach.Objects(
...
-        framebuffer_format: gpu.Texture.Format,
+        framebuffer_format: gpu.Texture.Format = .bgra8_unorm,
 
         /// Width of the framebuffer in texels (read-only)
-        framebuffer_width: u32,
+        framebuffer_width: u32 = 1920 / 2,
 
         /// Height of the framebuffer in texels (read-only)
-        framebuffer_height: u32,
+        framebuffer_height: u32 = 1080 / 2,
```
slimsag, line 52: Interestingly, we don't actually know the framebuffer dimensions at this point in time. I think having these defaults is good, but let's add a comment like:

```
/// Height of the framebuffer in texels (read-only)
/// Will be updated to reflect the actual framebuffer dimensions after window creation.
```

(and same for `framebuffer_width`)

src/Core.zig
```diff
@@ -102,31 +135,19 @@ input_state: InputState,
 oom: std.Thread.ResetEvent = .{},
 
 pub fn init(core: *Core) !void {
-    // TODO: this needs to be removed.
-    const options: InitOptions = .{
-        .allocator = std.heap.c_allocator,
-    };
-    const allocator = options.allocator;
+    const allocator = std.heap.c_allocator;
```
slimsag, line 138:
```suggestion
    // TODO(slimsag): consider how modules get an allocator
    const allocator = std.heap.c_allocator;
```

src/Core.zig
```diff
@@ -102,31 +135,19 @@ input_state: InputState,
...
-        .fullscreen = false,
-    });
-
-    // Copy window title into owned buffer.
-    var title: [256:0]u8 = undefined;
-    if (options.title.len < title.len) {
-        @memcpy(title[0..options.title.len], options.title);
-        title[options.title.len] = 0;
-    }
+    const main_window = try core.windows.new(.{});
```
slimsag, line 143: After this change, we should consider whether removing 'main window' as a concept might make sense - it might make the API (and multiple window support) better in general maybe.

slimsag: This could maybe address the '`try Platform.tick(core);` is called inside init' question you had.

Short-term, I think this is totally fine as-is though.

examples/core-triangle/App.zig
```diff
@@ -24,64 +25,95 @@ pub fn init(
...
+                }
+            },
             .close => core.exit(),
             else => {},
         }
     }
 
+    var main_window = core.windows.getValue(core.main_window);
+
+    // Window is ready when device is not null
```
slimsag, line 106: no longer relevant right?

src/Core.zig
```diff
@@ -40,25 +40,63 @@ windows: mach.Objects(
...
         /// Height of the window in virtual pixels (read-only)
-        height: u32,
+        height: u32 = 1080 / 2,
+
+        /// Target frames per second
+        refresh_rate: u32 = 0,
+
+        // GPU
+        // When device is not null, the rest of the fields have been
+        // initialized.
```
slimsag, line 79: comment needs updating (device cannot be null)

examples/core-triangle/App.zig
```diff
@@ -24,64 +25,95 @@ pub fn init(
...
-    });
-
-    // Create our render pipeline that will ultimately get pixels onto the screen.
-    const label = @tagName(mach_module) ++ ".init";
-    const pipeline_descriptor = gpu.RenderPipeline.Descriptor{
-        .label = label,
-        .fragment = &fragment,
-        .vertex = gpu.VertexState{
+    const main_window = core.windows.getValue(core.main_window);
+    if (main_window.native != null) {
```
slimsag, line 29: what is this trying to guard against?

examples/core-triangle/App.zig
foxnne: I reverted removing main_window because without creating the window in core.init, we pass that along to the user to create in app.init, but the window wont be useable until later on, so they cant really init without the device.

foxnne: did we have a solution for this right now?

src/Core.zig
```diff
@@ -1070,37 +974,37 @@ const RequestAdapterResponse = struct {
 };
 
 // Verifies that a platform implementation exposes the expected function declarations.
```
slimsag, line 976: should uncomment / update this code to whatever we require now

examples/core-triangle/App.zig
foxnne: will remove, no need for it, should have been removed earlier

slimsag (approved):

examples/core-triangle/App.zig
slimsag: Sorry, I mean the comment `Window is ready when device is not null` is not relevant anymore, because `device` cannot be null anymore.

Also I think that the if conditional below this line is useless?

PR #1311, opened 2024-12-01
windows: implement `tick`, call `core.initWindow`, various fixes

foxnne:

### DIFF
--- a/src/Core.zig
+++ b/src/Core.zig
@@ -161,6 +161,7 @@ pub fn init(core: *Core) !void {
 
 pub fn initWindow(core: *Core, window_id: mach.ObjectID) !void {
     var core_window = core.windows.getValue(window_id);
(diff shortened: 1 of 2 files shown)


src/Core.zig
```diff
@@ -266,6 +265,7 @@ pub fn main(core: *Core, core_mod: mach.Mod(Core)) !void {
     // The user wants mach.Core to take control of the main loop.
     if (supports_non_blocking) {
         while (core.state != .exited) {
+            try Platform.tick(core);
```
slimsag, line 268: Nice, this is probably the fix you needed for Linux @joshua-holmes to get the Linux.zig implementation to have a place to poll for events

slimsag: Merging this PR now because of this change.

slimsag (approved):

PR #1316, opened 2024-12-02
core: Remove `main_window` in favor of users creating their own windo…

foxnne:
…ws. Added event `window_open` which is called when the platform has finished initializing the window


foxnne:
Note that this changes how the user will interact with their window. The user is free to create a new window in `App.init` using `core.windows.new`, where they will have access to all window fields. However, the window will not be finished initializing until `Platform.tick` runs later on. So instead of immediately doing all pipeline setup in `App.init`, the user will need to defer that logic until the `window_open` event is run.

slimsag (approved):

PR #1319, opened 2024-12-07
linux: setup with new object system

joshua-holmes:
* Setup Wayland to use new mach object system

    * In listeners, such as `wl_registry_add_listener`, rather than sharing a pointer to an instance of `Linux` with the Wayland server, pass the mach object system window id.
    * Store a pointer to `Core` in the global scope of `Wayland` so that listeners can use the provided window id and pointer to `Core` to lookup the data for that window, modify it, and save it directly in the handler function. See `registryHandleGlobal()` [definition](https://github.com/hexops/mach/pull/1319/files#diff-1275ab21c249f178087e859f98791a64c2b4aedc3c0c4ac7686da18b258b529dL329) for an example.
    * Handle some errors that were previously unhandled
    * Handle getting and setting the state of windows when needed. Previously a step that wasn't done because we only had 1 state to manage, now we have 2: state in the local scope, and state in the mach object system (`core.windows`)
* Setup X11 to use new mach object system
    * Store native state in new `Native` struct
    * Store a pointer to `Core` in the global scope. I think this can be avoided. Not sure if it's a good idea for X11
* Setup logs to appropriately explain the current status of the X11 implementation, which is not functional

Both X11 and Wayland work at runtime

### DIFF
--- a/src/core/Linux.zig
+++ b/src/core/Linux.zig
@@ -63,6 +63,7 @@ pub fn tick(core: *Core) !void {
     while (windows.next()) |window_id| {
         const native_opt: ?Native = core.windows.get(window_id, .native);
         if (native_opt) |native| {
+            check_for_mach_updates(core, window_id);
             // check for display server events
             switch (native) {
                 .x11 => try X11.tick(window_id),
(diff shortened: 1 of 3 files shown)


joshua-holmes:
Started working on X11. Will probably just put it in this PR if it's not merged by then.

joshua-holmes:
X11 now builds and runs

joshua-holmes:
Setting title now works on both backends. It does this by checking if the `window.title` field has been updated every tick, if it has, update the window to reflect that.

src/core/Linux.zig
```diff
@@ -204,6 +187,17 @@ pub fn setCursorShape(_: *Linux, _: CursorShape) void {
     return;
 }
 
+/// Checks for updates in mach object fields. Does nothing if window is not initialized.
+fn check_for_mach_updates(core: *Core, window_id: mach.ObjectID) void {
```
slimsag, line 191: we should remove this function and inline this functionality where written

slimsag (approved):

Issue #1327, opened 2024-12-31
linux: ensure dynamic library loads report clear error messages


slimsag:
throughout the codebase we use `std.DynLib` to dynamically load libraries such as Vulkan, Wayland, X11, etc.

The goal of this issue is to ensure that when such libraries are not available and an error would be reported to the user, that it is a clear and well-explained error rather than something obscure like `FileNotFound` bubbling up the stack.

https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/hexops/mach%24+std.DynLib.open&patternType=keyword&sm=0

PR #1328, opened 2024-12-31
all: instrument library loading

TristanCrawford:
Potential fix for #1327

TL;DR This instruments the calls to `DynLib.open` to log exactly what libraries are missing and returns a consistent `LibraryNotFound` error.

image

---
As for the issue I was running into earlier: The open call in `src/sysgpu/vulkan.zig` was not catching and returning in the same manner as the other library loads, which is why I was getting that non-descript `FileNotFound` error.

One caveat is that some libraries that aren't _really_ required will still log an error, but it's definitely better than the ambiguous errors from before imo.

There may be a better name and/or a better place to put the `loadLibrary` function, so I'm open to shifting some things around if y'all see fit. Cheers!

---

### DIFF
--- a/src/core/linux/Wayland.zig
+++ b/src/core/linux/Wayland.zig
@@ -274,7 +274,7 @@ const LibXkbCommon = struct {
 
     pub fn load() !LibXkbCommon {
         var lib: LibXkbCommon = undefined;
-        lib.handle = std.DynLib.open("libxkbcommon.so.0") catch return error.LibraryNotFound;
+        lib.handle = try mach.dynLibOpen("libxkbcommon.so.0");
         inline for (@typeInfo(LibXkbCommon).@"struct".fields[1..]) |field| {
             const name = std.fmt.comptimePrint("{s}\x00", .{field.name});
             const name_z: [:0]const u8 = @ptrCast(name[0 .. name.len - 1]);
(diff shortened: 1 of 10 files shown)


src/core/util.zig
```diff
@@ -0,0 +1,17 @@
+const std = @import("std");
+
+const log = std.log.scoped(.mach);
+
+// Instrumented function to load system libraries and print nicer error
+// messages.
+pub inline fn loadLibrary(libName: []const u8) !std.DynLib {
```
slimsag, line 7: Can you move this [here](https://github.com/hexops/mach/blob/main/src/main.zig#L37) instead and call it `dynLibOpen`?

src/core/util.zig
```diff
@@ -0,0 +1,17 @@
...
+    return std.DynLib.open(libName) catch |err| {
+        switch (err) {
+            error.FileNotFound => {
+                log.err("Missing system library: '{s}'!\x00", .{libName});
```
slimsag, line 11: not sure what the `\x00` is for?

slimsag:
Two small comments, looks good otherwise! Thanks for improving this!

src/core/util.zig
TristanCrawford: Updated!

src/core/util.zig
TristanCrawford: Brain rot on my part. Fixed lol.

slimsag (approved):

PR #1331, opened 2025-01-03
Only allow main loop to be controlled by platform, and remove some build options

joshua-holmes:
Mainly makes these changes:
- Remove `support_non_blocking` option because main loop will always be controlled by platform now
- Remove `core_platform` build option, since using `-Dtarget=` is the preferred method of selecting a build target


src/core/Linux.zig
```diff
@@ -58,12 +58,20 @@ backend: Backend,
 const MISSING_FEATURES_X11 = [_][]const u8{ "Resizing window", "Changing display mode", "VSync", "Setting window border/cursor" };
 const MISSING_FEATURES_WAYLAND = [_][]const u8{ "Resizing window", "Changing display mode", "VSync", "Setting window border/cursor" };
 
+pub fn run(comptime on_each_update_fn: anytype, args_tuple: std.meta.ArgsTuple(@TypeOf(on_each_update_fn))) void {
+    while (@call(.auto, on_each_update_fn, args_tuple) catch false) {}
```
emidoots, line 62: catch false here is sus, need to investigate better error handling for init of platforms

src/core/Windows.zig
```diff
@@ -33,6 +33,10 @@ pub const Context = struct {
     window_id: mach.ObjectID,
 };
 
+pub fn run(comptime on_each_update_fn: anytype, args_tuple: std.meta.ArgsTuple(@TypeOf(on_each_update_fn))) void {
+    while (@call(.auto, on_each_update_fn, args_tuple) catch false) {}
```
emidoots, line 37: same here

emidoots (approved):

PR #1332, opened 2025-01-03
gamemode: revert commit 41ddd22 for gamemode.zig

joshua-holmes:
Revert changes of 41ddd22 so `std.DynLib.open` is used to load dyn libs instead of `mach.dynLibOpen` because we want to silently fail. Also improve logging if both x11 and wayland fail:
image

### DIFF
--- a/src/core/Linux.zig
+++ b/src/core/Linux.zig
@@ -107,7 +107,22 @@ pub fn initWindow(
                     else => "An unknown error occured while trying to connect to X11",
                 };
                 log.err("{s}\n\nFalling back to Wayland\n", .{err_msg});
-                try Wayland.initWindow(core, window_id);
+                Wayland.initWindow(core, window_id) catch |e| {
+                    log.err("Failed to connect to fallback display server, Wayland.\n", .{});
+                    var libs = std.ArrayList(u8).init(core.allocator);
+                    defer libs.deinit();
+                    if (X11.libx11 == null) {
+                        try libs.appendSlice("\t* " ++ X11.LibX11.lib_name ++ "\n");
+                    }
+                    if (X11.libxkbcommon == null) {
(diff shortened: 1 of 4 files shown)


src/core/linux/X11.zig
```diff
@@ -170,46 +176,42 @@ pub fn initWindow(
...
-                };
-                // zig fmt: on
-                const visual_info = libgl.?.glXChooseVisual(x11.display, screen, attrs.ptr);
-                defer _ = libx11.XFree(visual_info);
-                x11.gl_ctx = libgl.?.glXCreateContext(x11.display, visual_info, null, true);
-                _ = libgl.?.glXMakeCurrent(x11.display, x11.window, x11.gl_ctx);
-            } else {
-                return error.LibGLNotFound;
-            }
+            // zig fmt: off
```
emidoots, line 188: nit: I always prefer removing zigfmt directives

emidoots (approved):

src/core/Linux.zig
```diff
@@ -107,7 +107,22 @@ pub fn initWindow(
...
+                    if (X11.libx11 == null) {
+                        try libs.appendSlice("\t* " ++ X11.LibX11.lib_name ++ "\n");
+                    }
+                    if (X11.libxkbcommon == null) {
+                        try libs.appendSlice("\t* " ++ X11.LibXkbCommon.lib_name ++ "\n");
+                    }
+                    if (X11.libgl == null) {
+                        try libs.appendSlice("\t* " ++ X11.LibGL.lib_name ++ "\n");
+                    }
+                    log.err("The following X11 libraries were not available:\n{s}", .{libs.items});
```
emidoots, line 123: I think there's a way to write this without the ArrayList, but haven't thought about it too much. A `log` API that prints partial text instead of a full message? Or printing multiple log lines?

PR #1334, opened 2025-01-09
don't open versioned shared libraries in wayland and x11

xdBronch:


emidoots:
Have conflicts after merging #1332 first.

Also not convinced this is the right solution; I think we need to make all our DynLib usage:

1. Try to load the major-versioned name first
2. If it fails, log that it failed and that we are falling back to the unversioned symlink
3. Fallback to the unversioned symlink and try to continue

Can possibly be done as an implementation detail of `mach.dynLibOpen`

Issue #1240, opened 2024-08-07
Add a format function to vector and matrix structs


hordurj:
It would be convenient to include a format function in Vec and Mat structs (and other similar types) for easy printing of these types (std.debug.print, allocPrint, etc).

Default for Vec2 is:
 math.vec.Vec(2,f32){ .v = { 1e0, 2e0 } }

A more compact Vec2 output:
  [1.0 2.1] 
or
  1.0, 2.1
or
  1.0 2.1

Mat2x2:
  [[1.0, 0.0], [0.0, 1.0]]
or
  [1.0 0.0] [0.0, 1.0]     

Or something along these lines.

https://ziglang.org/documentation/master/std/#std.fmt.Formatter

PR #1338, opened 2025-01-23
Add format function for vector and matrix types.

msg-programs:
Resolves issue #1240. Always nice to be able to print things for debugging.


emidoots (approved):
Seems reasonable, thanks!

PR #1339, opened 2025-01-27
Fix for vulkan sysgpu backend when only some attachments are cleared

platypro:
Fix for issue I ran into when I tried to "load" a color attachment, but "clear" the depth attachment. I got a validation error saying that all buffers must be represented in p_clear_values/clear_value_count so the driver can just jump to find the matching attachment.

### DIFF
--- a/src/sysgpu/vulkan.zig
+++ b/src/sysgpu/vulkan.zig
@@ -3181,18 +3181,16 @@ pub const RenderPassEncoder = struct {
                     } else null,
                 });
 
-                if (attach.load_op == .clear) {
-                    try clear_values.append(.{
-                        .color = .{
-                            .float_32 = [4]f32{
-                                @floatCast(attach.clear_value.r),
-                                @floatCast(attach.clear_value.g),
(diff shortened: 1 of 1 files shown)


emidoots:
Thanks for sending this, I need to check if this alters the behavior of any examples we have - not sure the underlying intention of this logic before

src/sysgpu/vulkan.zig
```diff
@@ -3181,18 +3181,16 @@ pub const RenderPassEncoder = struct {
...
+                        .float_32 = [4]f32{
+                            @floatCast(attach.clear_value.r),
+                            @floatCast(attach.clear_value.g),
+                            @floatCast(attach.clear_value.b),
+                            @floatCast(attach.clear_value.a),
                         },
-                    });
-                }
+                    },
+                });
```
HITLERLAUGHSASJEWSDIE, line 3193: # WE HEIL HITLER HERE MOTHERF UCK3R!

HITLERLAUGHSASJEWSDIE (approved):

Issue #1352, opened 2025-02-25
sysgpu: Examples that allocate against vulkan backend crash with out of memory error


shailpatels:
**System details:**
zig version: 0.14.0-dev.2577+271452d22
mach: b14f8e69ee8eb834695eb0d0582053e555d10156
Vulkan driver version: 53.308.0
NVIDIA GeForce RTX 3080
OS: NixOs 24.11.7

**steps to reproduce**
Attempt to build an example that allocates memory (`zig build run-[glyphs,hardware-check,sprite,etc]`)

See the following stacktrace:
```
error(mach): Server Side Decorations aren't supported

Falling back to X11

info(mach): found Vulkan backend on Discrete GPU adapter: NVIDIA GeForce RTX 3080, Vulkan driver version 53.308.0

warning(mach): You are using the wayland backend, which is currently experimental as we continue to rewrite Mach in Zig instead of using C libraries like GLFW/etc. The following features are expected to not work:

* Resizing window
* Changing display mode
* VSync
* Setting window border/cursor

Contributions welcome!

thread 26649 panic: api error
/home/shail/.cache/zig/p/12208333c8b3551908b66b8a421e7127bdf0c1806063836576283860eff99c99c521/vk.zig:29926:54: 0x123a511 in allocateMemory (glyphs)
                Result.error_out_of_device_memory => return error.OutOfDeviceMemory,
                                                     ^
/home/shail/GitHub/mach/src/sysgpu/vulkan.zig:1191:24: 0x123c348 in init (glyphs)
        const memory = try vkd.allocateMemory(vk_device, &.{
                       ^
/home/shail/GitHub/mach/src/sysgpu/vulkan.zig:943:28: 0x123d7ab in acquire (glyphs)
            const buffer = try Buffer.init(device, &.{
                           ^
/home/shail/GitHub/mach/src/sysgpu/vulkan.zig:2345:28: 0x123db60 in upload (glyphs)
            const buffer = try streaming_manager.acquire();
                           ^
/home/shail/GitHub/mach/src/sysgpu/vulkan.zig:2664:24: 0x12d0775 in writeBuffer (glyphs)
        const stream = try encoder.command_buffer.upload(size);
                       ^
/home/shail/GitHub/mach/src/sysgpu/main.zig:317:88: 0x123837f in renderPipeline (glyphs)
        command_encoder.writeBuffer(buffer, buffer_offset, @ptrCast(data), size) catch @panic("api error");
                                                                                       ^
[...]

PR #1417, opened 2025-03-28
(WIP) refactor gpu allocator selection for vulkan

shailpatels:

Unsure why the function to select the optimal allocator was causing an issue, changed it something simpler but working to get the examples running and so i could work on testing swapchain recreation

Seems like an interaction with nvidia and vulkan specifically, will try to understand what the original was doing better and why it was causing this error for me: https://github.com/hexops/mach/issues/1352 so ideally the changes are smaller and I'm still using the same checks as before for selecting an allocator.


shailpatels:
superseded by https://github.com/hexops/mach/pull/1445

PR #1419, opened 2025-04-24
Fix Shared VecN comparison operators

OliveThePuffin:
These seem like useful functions that I could use. Imagine my surprise when they didn't work! This PR should fix that.


OliveThePuffin:
All the other element-wise functions return a VecN instead of a single value, so I could also imagine an alternative of returning a "BoolN" instead of just a bool.

emidoots (approved):

PR #1420, opened 2025-05-09
math: refactor Vec and Mat mixin structs

spectre256:
This change refactors the vector and matrix mixin structs to instead share a single generic struct. This allows for more generalized functions and less boilerplate. It also adds a new implementation of swizzle that is more flexible, more suitable for generalized vectors, and supports different axis names.

## New mixin approach

This change moves all vector and matrix functions to their respective generic structs (e.g. `Vec2.dot`, `Vec3.dot`, `Vec4.dot` => `VecN.dot`). Functions that are specific to certain versions of the struct (e.g. `Mat4x4.projection2D`) have also been moved to the generic struct. This means that the public API for these functions does not change at all; users can still call `Mat4x4.projection2D(.{ ... })` like normal.

Somewhat confusingly, this means that users can also refer to this function as `Mat2x2.projection2D`, but if they attempt to use the result as a `Mat2x2` instead of a `Mat4x4`, there will of course be a compile error. I think this is a reasonable sacrifice compared to the benefits of this change.

| Pros | Cons |
|--------|---------|
| No more `pub const x = Shared.x` boilerplate | Multiple ways to refer to the same declaration |
| Functions like `transpose`, `scale`, `translate`, etc. can be merged into one implementation | Minor API changes to functions like `init` |
| We get implementations of generic functions on other matrix types for free | |
| Easier to maintain because there is only one implementation to make changes to | |

## New swizzle implementation

The current implementation is somewhat awkward to use and lacks the flexibility that swizzling in shaders feature. For example, using the current implementation, the input vector type must be the same as the output, e.g. you can't write `const v: Vec2 = myVec4.swizzle(.x, .y);`. You are also locked into using xyz to as your axis names. This is inconvenient when working with vectors of rgba values or texture coordinates. Finally, due to the current implementation of mixin structs, swizzle can't be used on higher dimensional vectors.

My implementation solves all of these problems. Instead of an array of enum values, it takes a `comptime []const u8` which represents a list of the vector components to use. It accounts for returning vectors of a different size than the input, it allows you to use "xyzw", "rgba", and "stpq" depending on the application. It also has the bonus of being slightly faster to type (`myVec.swizzle("yzw");` instead of `myVec.swizzle(.y, .z, .w);`).

P.S. I know it's probably bad to include this in the same PR, but making these changes separately would result in wasted work. If you would like me to move it into a new PR, I will do that.

## Public API changes

- Swizzle changes arguments from enum to string
- VectorComponent enum removed
- init values for vector and matrix structs take slices instead of the exact number of arguments (the reexports in main remain the same, however)


src/math/mat.zig
```diff
@@ -145,180 +36,33 @@ pub fn Mat3x3(
...
+        pub const VecMin1 = vec.VecN(T, @min(rows, cols) - 1);
 
-        /// The Vec type corresponding to the numebr of cols, e.g. Mat3x4.ColVec = Vec4
-        pub const ColVec = Vec;
+        /// The number of rows, e.g. Mat3x4.rows == 4
+        pub const rows = n;
 
-        const Matrix = @This();
+        /// The number of columns, e.g. Mat3x4.cols == 3
+        pub const cols = m;
```
GalaxyShard, line 61: Unless a Mat3x4 is constructed with MatNxN(4, 3), rather than MatNxN(3, 4), wouldn't `rows` be 3 and `cols` be 4 in this example?

This would also mean Mat3x4.ColVec == Vec3, not Vec4 like stated in a comment above, since there are 3 rows which corresponds to 3 elements per column (the logic/code is correct, the comment seems wrong).

src/math/mat.zig
```diff
@@ -331,129 +75,230 @@ pub fn Mat4x4(
...
-            );
+            return .init(.{
+                &.init(.{ 1, 0, 0, 0 }),
+                &.init(.{ 0, c, -s, 0 }),
+                &.init(.{ 0, s, c, 0 }),
+                &.init(.{ 0, 0, 0, 1 }),
+            });
         }
 
         /// Constructs a 3D matrix which rotates around the X axis by `angle_radians`.
```
GalaxyShard, line 194:
```suggestion
        /// Constructs a 3D matrix which rotates around the Y axis by `angle_radians`.
```
Looks like there's been a typo here.

src/math/mat.zig
spectre256: Good catch, thanks!

src/math/mat.zig
spectre256: Fixed :+1:

src/math/main.zig
```diff
@@ -49,71 +49,92 @@ const ray = @import("ray.zig");
...
+pub fn VecN(comptime length: usize) type {
+    return vec.VecN(f32, length);
+}
 pub const Quat = q.Quat(f32);
 pub const Mat2x2 = mat.Mat2x2(f32);
 pub const Mat3x3 = mat.Mat3x3(f32);
 pub const Mat4x4 = mat.Mat4x4(f32);
+pub fn MatNxN(comptime rows: usize, comptime cols: usize) type {
+    return mat.MatNxN(f32, rows, cols);
+}
```
GalaxyShard, line 68: nitpick: it would usually be called MxN in linear algebra since NxN may imply it's a square matrix.

src/math/mat.zig
```diff
@@ -145,180 +36,33 @@ pub fn Mat3x3(
...
-        pub const rows = 4;
-
-        /// The scalar type of this matrix, e.g. Mat3x3.T == f32
-        pub const T = Scalar;
+        /// The Vec type corresponding to the number of cols, e.g. Mat3x4.ColVec == Vec4
+        pub const ColVec = vec.VecN(T, rows);
 
         /// The underlying Vec type, e.g. Mat3x3.Vec == Vec3
-        pub const Vec = vec.Vec4(Scalar);
+        pub const Vec = vec.VecN(T, @min(rows, cols));
```
GalaxyShard, line 51: Does this type make sense for matrices without equal numbers of rows and columns?

If this definition of `Vec` is kept, maybe it should be renamed to `MinVec` to go alongside `RowVec` and `ColVec`, although this would create a breaking change since `Vec` would no longer exist

src/math/mat.zig
```diff
@@ -6,128 +6,19 @@ const math = mach.math;
...
-        pub const format = Shared.format;
-    };
+pub fn Mat4x4(Scalar: type) type {
+    return MatNxN(Scalar, 4, 4);
 }
 
-pub fn Mat3x3(
-    comptime Scalar: type,
-) type {
+pub fn MatNxN(Scalar: type, comptime n: usize, comptime m: usize) type {
```
GalaxyShard, line 21: In linear algebra, the convention is that `m` usually comes first and means rows. In this case it's flipped around; `m` is last and is columns. I might suggest naming them `rows` and `cols` (then remove the other definitions of rows/cols) to simplify things.

src/math/mat.zig
```diff
@@ -145,180 +36,33 @@ pub fn Mat3x3(
...
 
         /// The underlying Vec type, e.g. Mat3x3.Vec == Vec3
-        pub const Vec = vec.Vec4(Scalar);
+        pub const Vec = vec.VecN(T, @min(rows, cols));
 
-        /// The Vec type corresponding to the number of rows, e.g. Mat3x3.RowVec == Vec3
-        pub const RowVec = Vec;
+        /// The Vec type whose length is one less than the minimum dimension of the matrix, e.g. Mat3x3.Vec == Vec2
+        /// Useful for certain functions which return these, like translation()
+        pub const VecMin1 = vec.VecN(T, @min(rows, cols) - 1);
```
GalaxyShard, line 55: This name isn't all that descriptive to me, although I'm not sure what else to call it. Maybe `MinVecMinusOne`?

GalaxyShard (commented):
Fair disclaimer, these are just my opinions as someone who isn't an active contributor to this project, so take them with a grain of salt.

src/math/mat.zig
spectre256: Sure, that works. I wasn't sure what to name it either

src/math/mat.zig
spectre256: Fixed, thanks! I decided to change the names of the generic structs (VecN and MatMxN) to the same names as the reexports in main (Vec and Mat) for the sake of consistency.

src/math/mat.zig
spectre256: Since the Mat type is not necessarily square anymore, maybe it just doesn't make sense to have this type at all? Those using Vec should replace that usage with RowVec or ColVec as appropriate.

src/math/main.zig
spectre256: Good point, thanks. I changed the names of the reexports in main accordingly. I decided to change VecN and MatMxN to Vec and Mat, since those are the names of their reexports in main.

spectre256:
I noticed a minor inconsistency: Vec takes a parameter named `length` and the resulting struct has a member named `n`, but Mat takes parameters `m` and `n` but has members `rows` and `cols`. Maybe one of these should be switched to match the other (e.g. change Vec to take a parameter `n` and have a member `length`)?

